PR #6, opened 2018-10-07
Maven without moving anything

Combustible:
Ok, I have another maven attempt that doesn't move any files around but still compiles a usable jar. This should work better!


Combustible:
This should work better than the last maven attempt. Only downside to this is you'd have to remember to update the <version>1.2</version> tag when you bump the version in plugin.yml for consistency.

JorelAli:
Wonderful! Works a charm, good job 👍

PR #19, opened 2018-10-20
Add exception handling & stacktrace for failed commands

Combustible:
Wrote some command code that had a NullPointerException - but had no way to figure out where it was (nothing in the log, just a failed command). 

This small snippet of code dumps a stacktrace to the log when a custom command throws an exception to help with debugging.


JorelAli:
I've always wanted to know how to fix that - thanks very much!

PR #29, opened 2018-12-01
Add dependency info to maven jar generation

Combustible:


Combustible:
This fixes not being able to use the provided jar to compile a dependent plugin because of missing brigadier dependency info.

PR #40, opened 2019-01-07
Exclude dependencies from final jar

MatrixTunnel:
This should be a non-breaking change and will reduce the jar size


JorelAli:
It doesn't necessarily reduce the jar size, since by default the dependency scope is "compile" and are not shaded into the jar, thus supplied at runtime. However, since this is good practice for a maven dependency and I'm always looking for ways to improve my knowledge of maven, I'll merge this.

JorelAli:
I see! Thanks for clarifying!

PR #61, opened 2019-07-26
Remove unnecessary folders/files

portlek:
It's not necessary.


JorelAli:
Thanks for pointing this out - I had planned to do so but totally forgot

PR #65, opened 2019-07-28
NMS jars from online mavn repo instead of local mvn repo

portlek:
People who want to contribute this project can't create their local mvn repo so we can get them from another nms mvn repo. Yes I know the downloadSpigot.sh but it didn't work on me so it's not stable.

### DIFF
--- a/1.13CommandAPI/pom.xml
+++ b/1.13CommandAPI/pom.xml
@@ -19,10 +19,6 @@
 
     <repositories>
         <repository>
-            <id>spigotmc-repo</id>
-            <url>https://hub.spigotmc.org/nexus/content/repositories/snapshots/</url>
-        </repository>
-        <repository>
             <id>minecraft-libraries</id>
             <name>Minecraft Libraries</name>
             <url>https://libraries.minecraft.net</url>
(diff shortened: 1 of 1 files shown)


portlek:
Also, why we need to download all of jar ? Three jars which are 1.13, 1.13.1 and 1.14 might be enough

JorelAli:
So, I see where you're coming from, but I do have good intention for using a local repository.

Firstly, due to the legal issues that exist with respect to the source code of the Minecraft Server, the creator of Spigot created a tool called the BuildTools, which builds the various versions of Spigot. Any developer/server owner that wants to get a copy of Spigot _should_ be using the BuildTools, as opposed to downloading redistributed copies of the Bukkit server.

Secondly, I know that `downloadSpigot.sh` _may_ not work for everyone _(I personally don't use it, which is completely ironic that I even made it)_, but nonetheless, using BuildTools manually isn't an insurmountable task, there's even [a page which states the commands you need to run](https://www.spigotmc.org/wiki/buildtools/#versions) to get the version you want.

Finally, I include all of the jars because the CommandAPI's compilation system uses the [SafeReflection](https://github.com/JorelAli/1.13-Command-API-SafeReflection) library. This is to ensure with 100% certainty that the code that the CommandAPI uses _actually works_ and can be tested for at compile time, as opposed to runtime. Due to the scale of the project, testing at runtime has been rendered less efficient _(as it's an API, you also have to create your own local projects which use the API just to check that it works)_.

-----

All of that said, you can literally just download them from [some website](https://getbukkit.org/download/spigot), and I don't think that's too big a deal.

PR #86, opened 2019-09-27
Update folder structure to follow standard Maven layout

ZiluckMichael:


ZiluckMichael:
This removes the need for using SafeReflection as all versions are handled natively. If you compare to other large frameworks that offer multiple versions, this is how they support them. It makes it significantly easier to maintain.

JorelAli:
Just glancing over the changes, I believe the need for SafeReflection is still required. SafeReflection is used for compile-time checks for reflective calls (e.g. getting a field named "a"). Without it, there is no way to determine for a specific version of Bukkit/Spigot whether certain fields exist/don't exist at compile time.

ZiluckMichael:
But with the changes in those commits you don't actually need to use reflection for anything. The wrappers you already have and the sub-module layout replaces them except for on the direct implementation layer.

JorelAli:
Ah yes, you're right. It seems that I overlooked that. Good job.

It looks good to merge, but I'd prefer to do my "regular testing" when it comes to merge requests, so in accordance with [this](https://github.com/JorelAli/1.13-Command-API#psa-v30-release-date), I'll plan to include it in the upcoming release, but for now will not do any further action.

I also plan to keep this pull request open until it has been merged _(otherwise I'd forget!)_

JorelAli:
@Michael-Ziluck Hi again, long time no see.
After taking some time out to work on other things, I've finally come back to the CommandAPI to release version 3.0. However, I've been unable to compile it properly with your proposed Maven layout, primarily because the spigot libraries (not to be confused with spigot API libraries), cannot be resolved.

Evidently, the spigot repository [here](https://hub.spigotmc.org/nexus/content/repositories/snapshots/org/spigotmc/) does not declare the `spigot.jar` files, only the `spigot-api.jar` files.

Unfortunately, I'll probably have to revert to the (very long and time consuming) technique of having to manually build the .jar files (which is compliant with the licenses of CraftBukkit) and linking them locally to the CommandAPI.

PR #95, opened 2020-04-24
Override the getMessage function

EnragedRabisu:
This preserves backwards compatibility for users who were catching a CommandSyntaxException and displaying the exception message by allowing them to only swap references of CommandSyntaxException to WrapperCommandSyntaxException without further code changes.

It might be worth overriding other Throwable methods to do the same thing.


JorelAli:
I see - indeed, this is a good pull request. I'll make sure to take a look at the other Throwable methods for v3.0 release

PR #104, opened 2020-06-29
added .gitignore, changed pom.xml, improved downloadSpigot.bat

ryuuta0217:
**I'm Japanese and my English is not good (I may use translations) please understand...**

Changes:
+ don't let git commit files that don't want... `(.gitignore)`
+ "org.spigotmc:spigot:1.15-R0.1-SNAPSHOT" is when run "downloadSpigot.bat", it doesn't add it to the local repository. The build fails, so I changed to a more reliable method.  `(1.13CommandAPI/commandapi-${mc-ver}/pom.xml , 1.13CommandAPI/commandapi-core/pom.xml)`
+ I changed the batch file a bit because I didn't want the garbage left in the root folder after I built Spigot. `(downloadSpigot.bat)`

Please let us know if you have any problems.
If you don't need this pull request or only need part of it, you may close it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,8 @@
 *.iml
 build-works/
 libs/
+bin
+.classpath
+.project
+.settings
+target
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+.idea/**
+*.iml
+build-works/
+libs/
--- a/1.13CommandAPI/.gitignore
+++ b/1.13CommandAPI/.gitignore
@@ -1,9 +1,0 @@
-bin
-libs
-.classpath
-.project
-.settings
-target
-spigotlibs/*
-.idea/
-*.iml
--- a/1.13CommandAPI/commandapi-1.13.1/pom.xml
+++ b/1.13CommandAPI/commandapi-1.13.1/pom.xml
@@ -16,6 +16,8 @@
             <groupId>org.spigotmc</groupId>
             <artifactId>spigot</artifactId>
             <version>1.13.1-R0.1-SNAPSHOT</version>
+            <scope>system</scope>
+            <systemPath>${basedir}/../../libs/spigot-1.13.1.jar</systemPath>
         </dependency>
         <dependency>
             <groupId>dev.jorel</groupId>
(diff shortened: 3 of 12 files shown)


JorelAli:
Hi! Currently, my downloadSpigot.bat file is misleading. (今、私 の downloadSpigot.bat は 紛らわしです).
When you run BuildTools.jar, the spigot jar is automatically added to your local machine's maven repository. As a result, this pull request is not needed.
However, I like your .gitignore file and will include it!.

ryuuta0217:
I understand!

However, in the current version of downloadSpigot.bat, the 1.15 dependency Probably not resolvable; change to 1.15.2 because it adds a dependency on 1.15.2 I think you need to do that.
So I think you need to change the pom.xml, I think.
(Probably this is why the build failed :()

Then, the ".gitignore" is supposed to be based on my batch file, and it's supposed to be It doesn't ignore Spigot or Bukkit folders. So, "libs/" and "build-works" are no longer needed and you can add
I think you'll need to add a "work/", "Bukkit/", "Spigot/", "BuildData/", "CraftBukkit/"

ありがとうございました！

<br>
<details>
<summary>自分の英語や翻訳の質が不安なので、日本語も残しておきます</summary>
わかりました！

しかしながら、現在のdownloadSpigot.batでは、1.15の依存関係はおそらく解決できません。1.15.2の依存関係が追加されるため、1.15.2に変更する必要があるように思います。
なので、pom.xmlを変更する必要があると私は思います。
(おそらくこれが原因でビルドに失敗しました :()

それから、この".gitignore"は、私のバッチファイルを基準に考えられており、Spigot や Bukkitフォルダーをignoreしません。なので、"libs/"や"build-works"は不要になり、追加で
"work/"、"Bukkit/"、"Spigot/"、"BuildData/"、"CraftBukkit/" の追加が必要になると思います。

ありがとうございました！
</details>

JorelAli:
I have updated the Linux/MacOS version (downloadSpigot.sh) which is a lot better (in this commit: ad6580707b7122bb1e607e30b48c47df80fc31c8). I hope to update downloadSpigot.bat later to be just as good.

少しの日本語ができます。私は日本語がじょうずじゃありません。でも、どういたしまして！

PR #147, opened 2020-10-08
4.3c release

JorelAli:


JorelAli:
Fixes a bug where functions wouldn't load properly because permission checks were stonking

PR #165, opened 2020-11-24
Untested quick draft of adding custom annotations.

jvdoorn:
This PR is a draft currently and attempts to support custom argument annotations.


JorelAli:
It has some interesting ideas, but I don't think I'll be using this method going forward.

Annotations are typically something that developers use, not write themselves. Asking developers to write their own annotations is unreasonable, considering how niche it is in the Java language, which makes it much less likely that developers will have a good understanding of how they work.

I think it's much better to instead have something like this:

```java
@Default
public static void myCommand(CommandSender sender, @ACustomArgument("someMethod") Object argument) {
    // Implementation here
}

@ACustomArgumentImplementation
public static CustomArgument<Object> someMethod() {
    // Returns a CustomArgument<Object>
}
```

It seems unintuitive, having the method name as a parameter, but because the CommandAPI's annotation system is performed at compile-time, it makes perfect sense, since undeclared methods are also caught at compile-time.

PR #166, opened 2020-11-26
CommandAPICommand Reorganization/Standardization

worldeva:
Originally CommandAPICommand was a bit disorganized:

- Grouped `set`, `get`, `with`, etc. commands together

- Added `//region` and `//endregion` to the areas to enable easier
navigation

- Marked `setPermission` as `deprecated` (or should `withPermission` be
`deprecated`?), as they do the same thing.

- Set commands now return `this` for easier chaining (more in line with
the `with` commands)

- Improved javadocs for the `set` commands to reflect what they actually do

Note: After doing this, a larger change may be in order: to overload the
with commands with equivalent methods and then to `deprecate` the `set`
commands, or to mark for internal usage.

### DIFF
--- a/CommandAPI/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPICommand.java
+++ b/CommandAPI/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPICommand.java
@@ -50,6 +50,9 @@ public class CommandAPICommand {
 		this.commandName = commandName;
 		this.isConverted = false;
 	}
+
+	//region With Commands (Adds/Subtracts traits)
+	//================================================================================
 	
 	/**
 	 * Applies a permission to the current command builder
@@ -143,7 +146,12 @@ public class CommandAPICommand {
 		this.subcommands.add(subcommand);
 		return this;
 	}
-	
+
+	//================================================================================
+	//endregion
+	//region Executes
+	//================================================================================
+
 	// Regular command executor 
 	
 	/**
@@ -187,9 +195,9 @@ public class CommandAPICommand {
 		this.executor.addResultingExecutor(executor);
 		return this;
 	}
-	
+
+
 	// Entity command executor
-	
 	/**
 	 * Adds an executor to the current command builder
 	 * @param executor A lambda of type <code>(Entity, Object[]) -&gt; ()</code> that will be executed when the command is run
@@ -199,7 +207,7 @@ public class CommandAPICommand {
 		this.executor.addNormalExecutor(executor);
 		return this;
 	}
-	
+
 	/**
 	 * Adds an executor to the current command builder
 	 * @param executor A lambda of type <code>(Entity, Object[]) -&gt; int</code> that will be executed when the command is run
@@ -209,9 +217,9 @@ public class CommandAPICommand {
 		this.executor.addResultingExecutor(executor);
 		return this;
 	}
-	
+
+
 	// Proxy command executor
-	
 	/**
 	 * Adds an executor to the current command builder
 	 * @param executor A lambda of type <code>(Entity, Object[]) -&gt; ()</code> that will be executed when the command is run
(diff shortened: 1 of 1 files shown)


CommandAPI/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPICommand.java
```diff
@@ -50,6 +50,9 @@ public CommandAPICommand(String commandName) {
 		this.commandName = commandName;
 		this.isConverted = false;
 	}
+
+	//region With Commands (Adds/Subtracts traits)
```
JorelAli, line 54: `//region` and `//endregion` is not a Java standard. Adding this in solely this one class introduces a complete inconsistency in code style with the rest of the project.

CommandAPI/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPICommand.java
```diff
@@ -295,7 +303,81 @@ public CommandAPICommand executesNative(NativeResultingCommandExecutor executor)
...
+	//================================================================================
+
+
+	/**
+	 * Sets the permissions required to run this command
+	 * @param permission the permission required to run this command
+	 * @return this command builder
+	 * @deprecated setPermission and withPermission function exactly the same
+	 */
+	public CommandAPICommand setPermission(CommandPermission permission) {
```
JorelAli, line 321: `setPermission` is not deprecated. Even though its functionally equivalent to `setPermission`, it has different semantics. `withPermission` is designed to be used by the builder, whereas `setPermission` is a setter, with its corresponding getter method `getPermission`

CommandAPI/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPICommand.java
```diff
@@ -392,26 +432,29 @@ public CustomCommandExecutor getExecutor() {
...
 	public boolean isConverted() {
 		return isConverted;
 	}
 
+	/**
+	 * @param isConverted
+	 * @todo Do we mark this as depricated? There are no javadocs for this so I'd assume its
+	 * for internal usage, but then shouldn't you mark as protected? I'm gonna leave this alone for now
+	 */
 	void setConverted(boolean isConverted) {
```
JorelAli, line 454: This is marked with default access (which is stricter than public, hence doesn't have to be documented.

CommandAPI/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPICommand.java
```diff
@@ -328,29 +402,11 @@ public void setPermission(CommandPermission permission) {
...
-	}
-
-	/**
-	 * Sets the requirements that must be satisfied to run this command
-	 * @param requirements the requirements that must be satisfied to run this command
-	 */
-	public void setRequirements(Predicate<CommandSender> requirements) {
-		this.requirements = requirements;
-	}
+	public Predicate<CommandSender> getRequirements() { return requirements; }
```
JorelAli, line 409: Inline methods is bad code style that doesn't fit with everything else in every other class in this project.

JorelAli:
This pull request has good intentions but misses a few key points:

- The code you are criticizing was entirely automatically generated by the Eclipse IDE to add getters and setters to the `CommandAPICommand` class, for the sole purpose of cloning an instance. 
- In hindsight, none of these getters nor setters should have been made, and instead it should have just implemented `Cloneable`, which is probably something I'll do for v6.0.

That being said, I'll include that as a new issue.

CommandAPI/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPICommand.java
worldeva: Hm... I think that that was automatically done by my ide; must of slipped by-

PR #204, opened 2021-05-15
Maybe this can fix the permission error?

portlek:


JorelAli:
Permission error has been fixed. Bash script doesn't want to support arrays right now.

Issue #193, opened 2021-04-30
Async suggestion support.


portlek:
**Please describe your suggestion**
Async support for overriding suggestion.

**Please supply examples of the desired inputs and outputs**
```java
new PlayerArgument("player").overrideSuggestions(sender -> {
  return CompletableFuture.supplyAsync(() -> {
    final var players = new ArrayList<String>();
    players.add("db-player-1");
    players.add("db-player-2");
    players.add("db-player-3");
    return players.toArray(new String[0]);
  });
});
```

JorelAli:
If my understanding is correct, you're trying to run the suggestions code independent of the main server thread. As far as I am aware with my limited asynchronous knowledge, it's not possible.

JorelAli:
Have you considered caching the result and simply passing the "completed" result to the suggestion provider?

portlek:
It's not a constant or cacheable value.

JorelAli:
I don't see how this could ever work in practice. When a player starts typing a command, it tells the server to request suggestions. When the server receives whatever packet telling it to generate suggestions, it must instantly respond with the list of suggestions - the server cannot delay this response. If the server wants to delay this response, the server has to block the main thread (which is the whole server).

I don't think this is possible for the CommandAPI.

JorelAli:
This is indeed possible! Massive kudos to the pull request from #249.

![asyncsuggestionscomingsoon](https://user-images.githubusercontent.com/4613171/144725566-83956bad-af4a-4811-9905-0cde879245e4.gif)

JorelAli:
Implemented for the next release (hopefully 6.6.0) in a few days time.

JorelAli:
Implemented in 7.0.0.

PR #249, opened 2021-11-27
Asynchronous Suggestions using CompletableFuture

469512345:
Brigadier deals with suggestions asynchronously (ish) using CompletableFuture, so this PR allows suggestions to be given as a CompletableFuture too.

I gave all of the methods the suffix 'C' as they have same erasure as the existing ones, but feel free to improve these names if approved.


JorelAli:
Wow, this looks very useful! I have to admit, I'm personally not very familiar with how the whole "completable future" mechanic works in Java, but this definitely looks like something that can be used well.

I'm currently fairly busy this weekend (and plan to get the 6.4.0 CommandAPI version released in the next few days in preparation for Minecraft 1.18's release), but after I've released that I'll make sure to come back to this!

469512345:
Yea that'd be brilliant. I was trying to generate suggestions using the results of other suggestions from brigadier, which requires the async nature like this.

CompletableFuture is basically Java's equivalent of JavaScript's promise api if that makes it any more intuitive.

JorelAli:
Ah okay, that helps me understand it much better, thanks!

JorelAli:
@469512345 Apologies if I'm doing something completely wrong here (if I am, I would love you to point me in the right direction - again, I've not really used completable futures in Java before!), but I haven't managed to get your completable future implementation to generate asynchronous suggestions.

To test this out, I implemented 3 commands:

- `/cmd5sec` which generates suggestions after 5 seconds. This was to simulate a long asynchronous task, such as a network request or a large database read task.
- `/cmdinstant` which generates suggestions "instantly" using the completable future method. This simulates a fast asynchronous task, such as a square root calculation of a number.
- `/cmdnormal` which generates suggestions normally without using an asynchronous method. This is a control method.

I used the following code to implement these commands:

```java
Argument stringArgumentReturnAfter5Seconds = new StringArgument("strarg").includeSuggestionsC(info -> {
	return CompletableFuture.supplyAsync(() -> {
		
		// Print wait for 5 seconds. We use the Bukkit scheduler to make it run on the main thread
		Bukkit.getScheduler().scheduleSyncDelayedTask(plugin, () -> Bukkit.broadcastMessage("Waiting for 5 seconds"));
		
		// Wait for 5 seconds. Simulates a long async task (e.g. network request)
		try {
			Thread.sleep(1000 * 5);
		} catch (InterruptedException e) {
			e.printStackTrace();
		}
		
		// Print done
		Bukkit.getScheduler().scheduleSyncDelayedTask(plugin, () -> Bukkit.broadcastMessage("Done!"));
		return new String[] { "hello", "world", "waddup" };
	});
});

Argument stringArgumentReturnInstantly = new StringArgument("strarg").includeSuggestionsC(info -> {
	return CompletableFuture.supplyAsync(() -> {
		return new String[] { "hello", "world", "waddup" };
	});
});

Argument stringArgumentNormal = new StringArgument("strarg").includeSuggestions(info -> {
	return new String[] { "hello", "world", "waddup" };
});

new CommandAPICommand("cmd5sec")
	.withArguments(stringArgumentReturnAfter5Seconds)
	.executes((sender, args) -> {
		Bukkit.broadcastMessage((String) args[0]);
	})
	.register();

new CommandAPICommand("cmdinstant")
.withArguments(stringArgumentReturnInstantly)
.executes((sender, args) -> {
	Bukkit.broadcastMessage((String) args[0]);
})
.register();

new CommandAPICommand("cmdnormal")
.withArguments(stringArgumentNormal)
.executes((sender, args) -> {
	Bukkit.broadcastMessage((String) args[0]);
})
.register();
```

I then ran these commands in-game, which displayed the following:

asyncsuggestions

As you can see, the asynchronous tasks did _not_ generate any suggestions, whereas the non-asynchronous suggestions did. I would expect the `/cmd5sec` command to generate suggestions (and for them to be displayed in the client) when the "Done!" message appeared and I would expect the `/cmdinstant` to generate suggestions just like in `/cmdnormal`.

I don't know if this is a Minecraft thing or a result of your implementation. I don't really want to merge in a feature which would cause this sort of interaction for CommandAPI users unless it works as intended.

Some insight into this would be much appreciated!

469512345:
I've been able to reproduce the behaviour demonstrated using version 6.5.2. Looking into it now.

469512345:
The issue was only occurring when including suggestions, not overwriting them (which is all I tested - my bad!).

The code for merging default and included suggestions wasn't suitable for asynchronous suggestions, this is my fault for missing this in the original commit. My apoligies! 
I've added a new commit which fixes this. All of your examples work fine now for me.

JorelAli:
This works as intended now. Lovely! 

asyncsuggestionscomingsoon

I'll work on a bit more manual testing for now and assuming everything goes well, I'll merge this into the CommandAPI (Please ignore CodeFactor's complaints, I'm aware my code is "too complex")

For what it's worth, this was actually a feature request opened earlier this year (#193) but I dismissed it because I thought it was impossible. I'm so glad this works!

Again, many thanks in advance for completing what I could not and improving the CommandAPI's functionality beyond what it was originally capable of.

JorelAli:
I'm happy with the result. I've merged this over to my [dev/dev branch](https://github.com/JorelAli/CommandAPI/commits/dev/dev) (Annoyingly, I can't tell GitHub "Mark this pull request as merged" without re-merging and re-applying various changes and screwing up my commit log, so apologies if this looks like I've not merged this pull request!) and will include this in the next CommandAPI release (hopefully 6.6.0) in a few days time.

I've renamed the methods to use the suffix "Async" instead of "C". Hopefully, I can write some good corresponding documentation to give suitable examples of this. Again, thanks very much for your contribution!

I'll continue update progress on this in the feature request #193.

PR #252, opened 2021-12-05
ArgumentSuggestions interface for providing suggestions

469512345:
Sorry for double submitting today, but I felt I had to write this while I remembered, and before the next version of CommandAPI comes out with the Async methods as I think this is a better way to handle suggestions in the long run.

This PR adds an `ArgumentSuggestions` interface which handles all suggestions operations.

This reduces the need for suggestions conversion code in `CommandAPIHandler`, and also many of the overridden methods in both `Argument` and `SafeOverrideableArgument`.

A `SafeSuggestions<S>` interface is also added which can be converted into an `ArgumentSuggestions` object using the string mapping function. (Unfortunately the `includeSafeSuggestions(SafeSuggestions<S>)` method has ambiguous signature with the `includeWithSafeSuggestions(Function<SuggestionInfo, S[]>)` so the name does have to be different!

Examples:

```java
new CommandAPICommand("weekday")
	.withArguments(
		new StringArgument("day").replaceSuggestions(
			ArgumentSuggestions.strings(
				"Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"
			)
		)
	).executes((sender, args) -> {
		sender.sendMessage("You chose " + args[0]);
	})
	.register();
```

weekday

Safe Suggestions

```java
new CommandAPICommand("coolbiomes")
	.withArguments(
		new BiomeArgument("biome").replaceSafeSuggestions(
			SafeSuggestions.suggest(
				Biome.BADLANDS, Biome.SMALL_END_ISLANDS
			)
		)
	).executes((sender, args) -> {
		sender.sendMessage("You chose " + args[0]);
	}).register();
```

coolbiomes

An argument which suggests as a complete brigadier command, using a completely custom suggestion provider.

```java
new CommandAPICommand("commandargument")
	.withArguments(new GreedyStringArgument("command").replaceSuggestions((info, builder) -> {

		String arg = info.currentArg();
		//Identify the position of the current argument
		int start;
		if(arg.contains(" ")) {
			//Current argument contains spaces - it starts after the last space and after the start of this argument.
			start = builder.getStart() + arg.lastIndexOf(' ') + 1;
		} else {
			//Input starts at the start of this argument
			start = builder.getStart();
		}
		//Parse command using brigadier
		ParseResults<?> parseResults = Brigadier
			.getCommandDispatcher()
			.parse(info.currentArg(),
				CommandAPIHandler
					.getInstance()
					.getNMS()
					.getCLWFromCommandSender(info.sender())
			);
		//Intercept any parsing errors indicating an invalid command
		for(CommandSyntaxException exception: parseResults.getExceptions().values()) {
			//Raise the error, with the cursor offset to line up with the argument
			throw new CommandSyntaxException(exception.getType(), exception.getRawMessage(), exception.getInput(), exception.getCursor() + start);
		}

		return Brigadier
			.getCommandDispatcher()
			.getCompletionSuggestions(parseResults)
			.thenApply((suggestionsObject) -> {
				//Brigadier's suggestions
				Suggestions suggestions = (Suggestions) suggestionsObject;
				return new Suggestions(
					//Offset the index range of the suggestions by the start of the current argument
					new StringRange(start, start + suggestions.getRange().getLength()),
					//Copy the suggestions
					suggestions.getList()
				);
			});
	})).executes((sender, args) -> {
		Bukkit.broadcastMessage(sender + " sent command: " + args[0]);
	}).register();
```

commandargumentnew

This command argument is a cool idea. It currently doesn't work properly with exceptions, but it's just an example of what can be done with a little bit of playing around.

So much more flexibility is given with this suggestion provider, and I do believe it would make maintaining the arguments much easier with fewer method overloads required. It feels like a natural progression from my last PR which first allowed suggestions via CompletableFuture.


JorelAli:
I really like this improvement. It makes the CommandAPI's code cleaner and makes suggestions more powerful. The command argument suggestion is exceptionally impressive.

The code works, asynchronous suggestions work as well. I'm happy to merge this over, except I'll merge this to my dev branch as opposed to master.

PR #253, opened 2021-12-06
Building a command using a Tree-like syntax

469512345:
This PR adds the `CommandTree` class, which provides an alternate way for building a command, using a tree structure. This aims to mirror Brigadier's command registration system, but using components of the CommandAPI.
 
This involved making lots of changes to the internals of `CommandAPICommand`, mainly moving parts of it into parent classes so that the `CommandTree` can inherit the same code. This is to hopefully make the code more maintainable for future changes. 

A `CommandMetaData` class has also been introduced to help store a command's metadata in a single object, which was required for the tree structure, but also massively reduces the number of argument for some methods in `CommandAPIHandler`.

A command can be built as follows

```java
new CommandTree("treeexample")
	//Set the aliases as you normally would 
	.withAliases("treealias")
	//Set an executor on the command itself
	.executes((sender, args) -> {
		sender.sendMessage("Root with no arguments");
	})
	//Create a new branch starting with a the literal 'integer'
	.then(new LiteralArgument("integer")
		//Execute on the literal itself
		.executes((sender, args) -> {
			sender.sendMessage("Integer Branch with no arguments");
		})
		//Create a further branch starting with an integer argument, which executes a command
		.then(new IntegerArgument("integer").executes((sender, args) -> {
			sender.sendMessage("Integer Branch with integer argument: " + args[0]);
		})))
	.then(new LiteralArgument("biome")
		.executes((sender, args) -> {
			sender.sendMessage("Biome Branch with no arguments");
		})
		.then(new BiomeArgument("biome").executes((sender, args) -> {
			sender.sendMessage("Biome Branch with biome argument: " + args[0]);
		})))
	.then(new LiteralArgument("string")
		.executes((sender, args) -> {
			sender.sendMessage("String Branch with no arguments");
		})
		.then(new StringArgument("string").executes((sender, args) -> {
			sender.sendMessage("String Branch with string argument: " + args[0]);
		})))
	//Call register to finish as you normally would
	.register();
```
Screen Recording 2021-12-06 at 23 03 25

This resulted in a small modification to the `Argument` to add the executes methods, but none of these return `Argument` so the below code will not compile. 

```java
new CommandAPICommand("badexample")
	//.executes returns `ArgumentTree` so this does not compile
	.withArguments(new LiteralArgument("node").executes((sender, args) -> {
		sender.sendMessage("This will not compile as .executes no longer returns an argument");
	}))
	.register();
```
This is to avoid potential confusing use cases like the above code. It does however mean that any argument operations (`withPermission`, `replaceSuggestions` etc) must be performed before any calls to `executes`; code will probably be more readable this way anyway.

Methods such as `withAliases`, and `withShortDescription` are only available on the root node to keep it simple as well.

Hopefully this will make it much easier to build complex commands with less code!


JorelAli:
Assuming this still allows the "old" syntax for command registration, I think I'll allow this.

When I originally designed the CommandAPI's command registration methods, I specifically chose NOT to use a tree structure like Brigadier because it can cause a huge hindrance to the readability of the code (especially when using tabs/spaces to indicate "what comes after what"). I had originally wanted a "similar to Bukkit command declarations" form of code rather than a tree structure.

That said, given that this is entirely optional for developers (and the normal command registration methods (I presume) work fine), giving more choice for developers is always a good thing.

I'll get back to this PR once I've had the time to play around with it myself.

469512345:
Yes the existing registration is untouched. This is an optional extra way to do it.
This tree implementation actually converts each executable into a CommandAPICommand internally for registering so it's really just a layer on top of the existing registration to make complex commands easier to define.

JorelAli:
Tested, everything works as expected. Happy with the result :)

469512345:
There is a bug actually, Line 42 in `ArgumentTree` should use a mutable list definition, `List.of` is immutable and causes problems for nested arguments beyond depth 2. `Lists.newArrayList` solves this.

JorelAli:
> There is a bug actually, Line 42 in `ArgumentTree` should use a mutable list definition, `List.of` is immutable and causes problems for nested arguments beyond depth 2. `Lists.newArrayList` solves this.

Fixed in commit 2b327650

PR #254, opened 2021-12-10
Fix a couple of bugs with the CommandTree implementation.

469512345:
Fixes a bug where `ArgumentTree#then` would prepend arguments onto the executions of the `ArgumentTree` argument, which would still be present if the same `ArgumentTree` is used elsewhere. Instead now the executions are duplicated with the prepended argument.

Delay the calculation of resulting `CommandAPICommand`s in a CommandTree to the `CommandTree#register()` call, instead of building this at each `CommandTree#then` call. This makes the timing of adding arguments less important.


JorelAli:
👍

Issue #298, opened 2022-07-01
CustomArgumentInfo.previousArgs is empty (should contain PlayerArgument)


BeefDev:
**CommandAPI version:** 8.4.1

**Minecraft version:** 1.18.2

**Shading the CommandAPI?** yes

**What I did:**

My plugin stores "profiles" (just an object in memory) for all players. Admins should be able to access profiles by other players, for that purpose I have two arguments. SenderOwnedProfileArgument, which gets the player who is the command sender and gets their profiles, which works perfectly. TargetOwnedProfileArgument, for the admin commands, which gets a previous arg, casts it to a player and follows the same procedure, this breaks.

This is the code for the custom argument which breaks: https://www.toptal.com/developers/hastebin/cubihezazu.java
This is the line where I add the arguments: `.withArguments(new PlayerArgument("target"), new TargetOwnedProfileArgument(this.plugin.getAPI().getPlayerList(), messages, "profile", 0).get())`
This is the command I run: /adminprofiles profile BeefDevClone a

The withArguments method is called on a subcommand (profile) which is later added to adminprofiles as a subcommand. BeefDevClone is supposed to be the player argument (name of my account), and a is the name of one BeefDevClone's profiles.

**What actually happened:**

After I executed the command I get a `Error in executing command adminprofiles profile BeefDevClone a - a<--[HERE]`. Printing inside the function which creates the argument (before line 35 in the code). Shows that the previousArgs array is empty, even though the player argument is before the custom one. This causes the call in line 35 to fail because the array is empty (I assume an ArrayIndexOutOfBoundsException is thrown somewhere.).

Sadly no error logs apart from this.

**What should have happened:**

The PlayerArgument is added to the array, the custom argument gets it and executes normally.

BeefDev:
After some testing, I discovered that a LiteralArgument is placed before other arguments. The loop which attempts to find all previous arguments stops on that one because it matches the node name. That command has a subcommand with the exact same name as the nodename of the argument. In place of that subcommand the LiteralArgument is placed. After reading the source of LiteralArgument, I believe that it is indeed there in place of the subcommand. Since the subcommand is "profile" and the nodename of the argument is "profile". Perhaps the loop thinks the LiteralArgument (which is the first one (index 0)) is the custom one, so it stops the loop there and returns the empty array.

[...]

PR #301, opened 2022-07-10
Avoid re-parsing arguments in CommandAPIHandler#parseArgument

willkroboth:
`CommanAPIHandler#parseArgument` is called in two places, `CommandAPIHandler#argsToObjectArr` and `CommandAPIHandler#generatePreviousArguments`. Previously, `parseArgument` called `generatePreviousArguments` to get the objects associated with the previous arguments in case they were needed. Since `generatePreviousArguments` goes back and calls `parseArgument`, this can cause arguments to be parsed multiple times like so:
```
argsToObjectArr -> I need to parse args [1, 2, 3]
1:
    parseArgument -> I need previous args for 1
        generatePreviousArguments -> There are none
    1 is parsed
2:
    pA -> I need previous args for 2
        gPA -> I need to parse arg 1
            pA -> I need previous args for 1
                gPA -> There are none
            1 is parsed
    2 is parsed
3:
    pA  -> I need previous args for 3
        gPA -> I need to parse args 1, 2
            pA -> I need previous args for 1
                gPA -> There are none
            1 is parsed
            pA -> I need previous args for 2
                gPA -> I need to parse arg 1
                    pA -> I need previous args for 1
                        gPA -> There are none
                    1 is parsed
            2 is parsed
    3 is parsed
```
Assuming parsing the same argument gives the same result, this is unnecessary. The repetition can be avoided by building a list of the previously parsed args and passing that directly to `parseArgument`. Since the two methods that call `parseArgument` keep track of these lists anyways, making this change is easy.
With this change, the situation from above now goes like this:
```
argsToObjectArr -> I need to parse args [1, 2, 3]
args: []
1:
    parseArgument -> I have previous args for 1
    1 is parsed
args: [1]
2:
    parseArgument -> I have previous args for 2
    2 is parsed
args: [1, 2]
3:
    parseArgument -> I have previous args for 3
    3 is parsed
args: [1, 2, 3]

return args
```

This change fixes issue #298 since `generatePreviousArgs` is no longer called when running a command, which assumed all arguments had unique names. `CustomArgumentsInfo#previousArgs` now correctly gives all previous arguments, even when an argument with the same nodeName as the CustomArgument exists.

### DIFF
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
@@ -341,7 +341,7 @@ public class CommandAPIHandler<CommandSourceStack> {
 		// Populate array
 		for (Argument<?> argument : args) {
 			if (argument.isListed()) {
-				argList.add(parseArgument(cmdCtx, argument.getNodeName(), argument, args));
+				argList.add(parseArgument(cmdCtx, argument.getNodeName(), argument, argList.toArray()));
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


JorelAli:
Thanks for the detailed analysis! I'll merge this over for now, but will stare at #298 a bit more in my own time (I want to do some testing with literal arguments because I don't think they're doing what I want)

PR #302, opened 2022-07-18
isLoaded method

MBDOFF:
Adds isLoaded method to CommandAPI.java to avoid getting "You've tried to call the CommandAPI's onLoad() method more than once!".


JorelAli:
Looks good to me, thanks :)

PR #303, opened 2022-07-20
Adds the possibility to access CustomCommandExecutor and to clear the executors.

MBDOFF:


JorelAli:
@MBDOFF Wait hold on, using Collections.emptyList passes an immutable list to the CustomCommandExecutor - subsequent calls to .add...Executor will fail

PR #319, opened 2022-08-04
Improve ListArgument

willkroboth:
The focus of this PR is to improve the ListArgument's suggestions. Using brigadier's `SuggestionBuilder#createOffset` items can be suggested one at a time, which makes it more clear what the next item is. If an argument is completed this code also suggests the deliminator, making it clear to the user how the command wants them to separate the arguments.

As a minor addition, the ListArgumentBuilder was given the method `withList(T... array)`, which automatically wraps an array into a list. I think this is cleaner on the developer side since they don't need to put `List.of()` themselves. Regardless, it's just a minor change, so you can keep it or leave it.

### DIFF
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/arguments/ListArgument.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/arguments/ListArgument.java
@@ -30,6 +30,7 @@ import java.util.Set;
 import java.util.function.Function;
 import java.util.regex.Pattern;
 
+import com.mojang.brigadier.LiteralMessage;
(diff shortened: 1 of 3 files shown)


JorelAli:
PR is very nice and greatly improves the user's experience with the ListArgument. It's important to ensure that users don't accidentally put a trailing comma at the end of their list (e.g. `/cmd item1, item2,`, because then the second item (`item2`) won't appear in the list. I think this is a suitable solution nonetheless - it solves the massive block of suggestions which overall is a better experience.

_(I meant to comment this before merging, but oh well)_

PR #322, opened 2022-08-07
Improvements to suggestions with tooltips

469512345:
This PR pretty much just adds more support for using formatted chat without ChatColor.RED concatenation hell. This is provided in places where brigadier accepts the `Message` type, which is tooltips and `WrapperCommandSyntaxException`. Some utility methods for working with tooltips on safe suggestions are also provided (see below).

Method overrides using strings are still provided for convenience.

Methods provided in `Tooltip.java`

```java
//Methods for creating a tooltip (the same also exist in StringTooltip)
	public static <S> Tooltip<S> of(S object, String tooltip) {
		return of(object, toMessage(tooltip));
	}

	public static <S> Tooltip<S> of(S object, Message tooltip) {
		return new Tooltip<S>(object, tooltip);
	}

	public static <S> Tooltip<S> of(S object, BaseComponent... tooltip) {
		return of(object, toMessage(tooltip));
	}

//Methods for creating messages from strings or bungee chat components
	public static Message toMessage(String string) {
		return new LiteralMessage(string);
	}

	public static Message toMessage(BaseComponent... components) {
		return CommandAPIHandler.getInstance().getNMS().componentsToMessage(components);
	}
```

The advantage of using this brigadier `Message` type is that the adventure chat api components implement `Message` directly so no conversion is necessary.

An NMS method called `componentsToMessage` is also provided which will convert bungee `BaseComponent[]` chat components into a native nms component which also implements `Message`.  Although the bungee chat api seems to be deprecated so supporting it might be necessary...

In addition to this, methods for generating tooltips using a mapping function are provided. This is mainly aimed at `SafeSuggestions` where it may be useful to programatically generate the tooltips. I would still argue that implementing `IStringTooltip` is preferable if possible as It follows OOP, but this is not possible for builtin types.

more methods provided in `Tooltip.java`

```java
	@SafeVarargs
	public static <S> Tooltip<S>[] generateStrings(Function<S, String> tooltipGenerator, S... suggestions) {
		return generate(tooltipGenerator, Tooltip::of, suggestions);
	}

	@SafeVarargs
	public static <S> Tooltip<S>[] generateMessages(Function<S, Message> tooltipGenerator, S... suggestions) {
		return generate(tooltipGenerator, Tooltip::of, suggestions);
	}

	@SafeVarargs
	public static <S> Tooltip<S>[] generateComponents(Function<S, BaseComponent[]> tooltipGenerator, S... suggestions) {
		return generate(tooltipGenerator, Tooltip::of, suggestions);
	}

//Base method for the above 3 methods
@SafeVarargs
	private static <S, T> Tooltip<S>[] generate(Function<S, T> tooltipGenerator, BiFunction<S, T, Tooltip<S>> tooltipWrapper, S... suggestions) {
		Tooltip<?>[] tooltips = new Tooltip<?>[suggestions.length];
		for(int i = 0; i < suggestions.length; i++) {
			S suggestion = suggestions[i];
			tooltips[i] = tooltipWrapper.apply(suggestion, tooltipGenerator.apply(suggestion));
		}
		return (Tooltip<S>[]) tooltips;
	}
```

The same 3 methods have been provided in `StringTooltip.java` for parity, but this is probably less useful than generic suggestions.

In short this PR aims to expand the control over tooltips and how they're formatted, while also providing better ways to work with them.


JorelAli:
This is a relatively large PR and plan to release 8.5.1 soon (due on Friday 12th at the latest). As such, I won't be able to get round to this for 8.5.1's release, but I'll certainly have a look in more detail for a 8.5.2 (or even a 8.6.0 if this PR is deemed large enough) release.

JorelAli:
I've updated the PR with the latest changes (8.5.1's release). I've begun drafting a review of this PR and will get back to you when I'm done. For all intents and purposes, the next version of the CommandAPI will be _8.6.0_.

JorelAli:
> The advantage of using this brigadier Message type is that the adventure chat api components implement Message directly so no conversion is necessary.

I might be missing something, but can you prove this for me? I've not been able to confirm that Adventure's components implement Message when using the following test code:

```java
StringTooltip.ofMessage("text", Component.text("hello"));
```

469512345:
> > The advantage of using this brigadier Message type is that the adventure chat api components implement Message directly so no conversion is necessary.
> 
> I might be missing something, but can you prove this for me? I've not been able to confirm that Adventure's components implement Message when using the following test code:
> 
> ```java
> StringTooltip.ofMessage("text", Component.text("hello"));
> ```

Apologies I was incorrect in stating this. I'm not sure how I got this into my head.

I've provided proper implementation for this in commit 4eb0e45

JorelAli:
Looks very promising. I should be able to have a test of this PR this time tomorrow or the day after and hopefully we can get this merged over!

commandapi-core/src/main/java/dev/jorel/commandapi/StringTooltip.java
```diff
@@ -28,8 +35,58 @@
...
+
+	/**
+	 * Constructs a StringTooltip with a suggestion and a string tooltip
+	 *
+	 * @param suggestion the suggestion to provide to the user
+	 * @param tooltip    the string tooltip to show to the user when they hover over the
+	 *                   suggestion
+	 * @return a StringTooltip representing this suggestion and tooltip
+	 */
+	public static StringTooltip ofString(String suggestion, String tooltip) {
```
JorelAli, line 50: I like the naming of these methods (e.g. `ofString`, `ofMessage` etc.), but we need to retain an `of(String, String)` method for backward compatibility. (I'm a big fan of `ofString()`, please keep this and implement a deprecated `of(String, String)` method as well)

commandapi-core/src/main/java/dev/jorel/commandapi/StringTooltip.java
```diff
@@ -28,8 +35,58 @@
...
+
+	/**
+	 * Constructs a StringTooltip with a suggestion and a formatted bungee text component tooltip
+	 *
+	 * @param suggestion the suggestion to provide to the user
+	 * @param tooltip    the formatted tooltip to show to the user when they hover over the
+	 *                   suggestion
+	 * @return a StringTooltip representing this suggestion and tooltip
+	 */
+	public static StringTooltip ofBungeeComponents(String suggestion, BaseComponent... tooltip) {
```
JorelAli, line 74: I'd rather the method name be called `ofBaseComponents` rather than `ofBungeeComponents` - a name reflecting the class name is definitely much more preferable. The CommandAPI doesn't support Bungee and the naming is misleading and possibly also confusing for Spigot developers that haven't used Bungee.

This also applies to `Tooltip.java` and the `ofBungeeComponents` and `generateBungeeComponents` methods

commandapi-core/src/main/java/dev/jorel/commandapi/Tooltip.java
```diff
@@ -34,13 +42,42 @@
...
+	/**
+	 * Constructs a <code>Tooltip&lt;S&gt;</code> with a suggestion and a tooltip
+	 *
+	 * @param <S> the object that the argument suggestions use
+	 * @param object the suggestion to provide to the user
+	 * @param tooltip    the tooltip to show to the user when they hover over the
+	 *                   suggestion
+	 * @return a <code>Tooltip&lt;S&gt;</code> representing this suggestion and tooltip
+	 */
+	public static <S> Tooltip<S> ofString(S object, String tooltip) {
```
JorelAli, line 77: We need an `of(S, String)` method for backwards compatibility

commandapi-core/src/main/java/dev/jorel/commandapi/Tooltip.java
```diff
@@ -90,7 +241,49 @@ public static <S> Tooltip<S>[] arrayOf(Tooltip<S>... tooltips) {
...
+	 *
+	 * This supports all forms of formatting including entity selectors, scores,
+	 * click &amp; hover events, translations, keybinds and more.
+	 *
+	 * Note: the bungee component api is deprecated, and the adventure text component api should be used instead
+	 *
+	 * @param components array of bungee text components
+	 * @return native minecraft message object which can be used natively by brigadier.
+	 */
+	@Deprecated
```
JorelAli, line 271: Deprecated method must have a corresponding `@deprecated` entry in the JavaDocs to point users to the alternative to using the deprecated method.

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
```diff
@@ -260,7 +261,18 @@ public static void reloadDatapacks() {
...
+	}
+
+	/**
+	 * Forces a command to return a success value of 0
+	 *
+	 * @param message Description of the error message, formatted as a brigadier message
+	 * @return a {@link WrapperCommandSyntaxException} that wraps Brigadier's
+	 *         {@link CommandSyntaxException}
+	 */
+	public static WrapperCommandSyntaxException fail(Message message) {
```
JorelAli, line 274: Adding `CommandAPI.fail` overloads for `BaseComponent[]` and Adventure's `Component` objects is preferred for completeness and consistency (and it's also currently being requested by #334!)

JorelAli (changes requested):

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
469512345: Added in d10438a

commandapi-core/src/main/java/dev/jorel/commandapi/Tooltip.java
469512345: Added in f500786

commandapi-core/src/main/java/dev/jorel/commandapi/Tooltip.java
469512345: Added in d2efc99

commandapi-core/src/main/java/dev/jorel/commandapi/StringTooltip.java
469512345: d4bc110

commandapi-core/src/main/java/dev/jorel/commandapi/StringTooltip.java
469512345: 997bbd9

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
469512345: I've observed some compilation errors with the compiler not being able to determine the intended method overload. This is what lead to `Tooltip#ofString` which removes the need for the compiler to determine the appropriate method.

Would it be better if`CommandAPI#fail` were deprecated and replaced with `CommandAPI#failWithString(String)`, `CommandAPI#failWithMessage(Message)`, `CommandAPI#failWithBaseComponents(BaseComponent[])`, and `CommandAPI#failWithAdventureComponent` to avoid this issue too?

469512345:
I've addressed all of the changes requested. I hope this is what was wanted.

The BaseComponent API is deprecated according to the sources downloaded in IntelliJ.
```java
/**
 * @deprecated BungeeCord Chat API has been deprecated in favor of Adventure API. For help with migrating your code, see <a href="https://docs.adventure.kyori.net/migration/bungeecord-chat-api.html">https://docs.adventure.kyori.net/migration/bungeecord-chat-api.html</a>
 */
```

Is it actually worth adding new support for an API which is already deprecated? Code-wise both bungee's `BaseComponent` and adventure's `Component` are serialised into JSON which is then turned into native components using `NMS#generateMessageFromJson(String json)` so the only real overhead from supporting `BaseComponent` is the additional method overloads.

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
JorelAli: > I've observed some compilation errors with the compiler not being able to determine the intended method overload. This is what lead to `Tooltip#ofString` which removes the need for the compiler to determine the appropriate method.
> 
> Would it be better if`CommandAPI#fail` were deprecated and replaced with `CommandAPI#failWithString(String)`, `CommandAPI#failWithMessage(Message)`, `CommandAPI#failWithBaseComponents(BaseComponent[])`, and `CommandAPI#failWithAdventureComponent` to avoid this issue too?

That's absolutely fine :)

JorelAli:
> The BaseComponent API is deprecated according to the sources downloaded in IntelliJ.

The BaseComponent API is deprecated _according to Paper_, which makes sense because paper has the Adventure API. For non-paper servers (i.e. plain old Spigot), the BaseComponent API is the only thing they have and in the Spigot API, it is _not_ deprecated. It is imperative that we add support for the BaseComponent API because the CommandAPI is a _Spigot_ plugin that is distributed on the _Spigot_ website - if there were some sort of Paper-specific plugin distribution platform, then it wouldn't be necessary.

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
469512345: 227f1d1

JorelAli (approved):

commandapi-core/src/main/java/dev/jorel/commandapi/Tooltip.java
```diff
@@ -90,7 +257,48 @@ public static <S> Tooltip<S>[] arrayOf(Tooltip<S>... tooltips) {
...
+	 * Converts a formatted adventure text component to a native minecraft text component which can be used natively by brigadier.
+	 *
+	 * This supports all forms of formatting including entity selectors, scores,
+	 * click &amp; hover events, translations, keybinds and more.
+	 **
+	 * @param component adventure text component
+	 * @return native minecraft message object which can be used natively by brigadier.
+	 */
+	public static Message messageFromAdventureComponent(Component component) {
+		return CommandAPIHandler.getInstance().getNMS().generateMessageFromJson(BukkitComponentSerializer.gson().serialize(component));
```
JorelAli, line 301: `net.kyori.adventure.platform.bukkit.BukkitComponentSerializer` is not suitable for Paper Minecraft servers!
You should use the `net.kyori.adventure.text.serializer.gson.GsonComponentSerializer` instead.

The `BukkitComponentSerializer` is for CraftBukkit servers (non-Spigot and non-Paper) and is an external library that must be downloaded for CraftBukkit servers. `GsonComponentSerializer` comes packaged with Paper servers and that's the recommended class to use (as `PaperComponents.gsonSerializer()` [is deprecated](https://jd.papermc.io/paper/1.19/io/papermc/paper/text/PaperComponents.html#gsonSerializer()))

JorelAli (changes requested):

JorelAli:
I've tested everything and everything's looking fantastic. There's just the one last point about `BukkitComponentSerializer` which should be `GsonComponentSerializer` and once that's completed we're good to merge!

469512345:
> I've tested everything and everything's looking fantastic. There's just the one last point about `BukkitComponentSerializer` which should be `GsonComponentSerializer` and once that's completed we're good to merge!

Done in 5ed2de3

JorelAli (approved):

JorelAli:
@469512345 This feature has already surpassed my expectations and has made a number of developers in my community happy, thanks very much for this PR! ❤️

469512345:
Myself included. Happy to contribute as always

PR #324, opened 2022-08-08
Throw CommandSyntaxExceptions when ListArgument input is wrong

willkroboth:
Adding one last change to the ListArgument I forgot for #319, this PR enforces the ListArgument's syntax by throwing CommandSyntaxExceptions. Previously, the ListArgument would ignore items that weren't in its list and move on. If the list didn't allow duplicates, it would also just add one of each item. Now, a CommandSyntaxException will be thrown in those cases.

I think the changes to the test cases best explain this:
```java
server.dispatchCommand(sender, "list cat, wolf, axolotl"); // normal list
server.dispatchCommand(sender, "list cat, wolf, axolotl, wolf"); // don't allow duplicates
server.dispatchCommand(sender, "list axolotl, wolf, chicken, cat"); // don't allow unknown items
```
This list command does not allow duplicates. The first command works. The second command complains that `wolf` is in the list twice. The third command complains that `chicken` is not a known argument.

I think this way of handling incorrect items in the list is better than the previous method. Before, if a player sent the command `list cat, wolf, axolotl, wolf` and didn't get any errors, they would expect that whatever the command did, it used `wolf` twice. No one would know that the second `wolf` was cut out of the list behind the scenes. Now, that player would get a command error saying `wolf <- Duplicate arguments are not allowed` and understand why the command isn't doing what they thought it should.

I did have to modify the ListArgument tests, and I'm not sure if I did that correctly. When an invalid command is sent I made it check that nothing got through using `assertNull(type.get());`. If there's a better way to check for a command error, you can put that in.

### DIFF
--- a/commandapi-plugin-test/src/test/java/ArgumentTests.java
+++ b/commandapi-plugin-test/src/test/java/ArgumentTests.java
@@ -1,10 +1,3 @@
-import static org.junit.jupiter.api.Assertions.assertArrayEquals;
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -48,6 +41,8 @@ import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.serializer.gson.GsonComponentSerializer;
 import net.md_5.bungee.api.chat.BaseComponent;
 import net.md_5.bungee.chat.ComponentSerializer;
+
+import static org.junit.jupiter.api.Assertions.*;
 
 /**
  * Tests for the 40+ arguments in dev.jorel.commandapi.arguments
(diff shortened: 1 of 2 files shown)


commandapi-plugin-test/src/test/java/ArgumentTests.java
```diff
@@ -396,18 +389,18 @@ public void executionTestWithListArgument() {
...
-		server.dispatchCommand(sender, "listfunc cat, wolf, axolotl");
-		server.dispatchCommand(sender, "listfunc cat, wolf, axolotl, chicken");
-		server.dispatchCommand(sender, "listfunc axolotl, wolf, chicken, axolotl");
-		server.dispatchCommand(sender, "listfunc axolotl, wolf, chicken, axolotl, " + sender.getName());
-	
-		assertEquals(List.of("cat", "wolf", "axolotl"), type.get());
+
+		server.dispatchCommand(sender, "listfunc cat, wolf, axolotl"); // normal list
+		server.dispatchCommand(sender, "listfunc cat, wolf, axolotl, wolf"); // don't allow duplicates
+		server.dispatchCommand(sender, "listfunc axolotl, wolf, chicken, cat"); // don't allow unknown items
```
JorelAli, line 395: In the tests, it's better to use `assertThrows` here. I'd rather have a valid `assertThrows` that actually throws rather than the `assertNull` at the bottom which could mean anything

JorelAli: You can use the `server.dispatchThrowableCommand` method which wraps commands via Brigadier to let you test whether commands fail or not, for example:
```java
assertThrows(CommandSyntaxException.class, () -> assertTrue(server.dispatchThrowableCommand(sender, "listfunc cat, wolf, axolotl, wolf")));
```

JorelAli:
I've tested the main code (i.e. not the automated tests) and it works nicely. The one minor issue I have with it is I don't like the error message "Incorrect argument for command" - I think a more specific error message would be better, for example "List item <item> is not a valid list entry".

JorelAli:
Awesome, thanks very much! I clearly had no idea how to put that error message into reasonable English, but you managed to do so with no issues :)

PR #326, opened 2022-08-16
Share and simplify ChatPreviewHandler code

willkroboth:
The focus of this PR is the added class `NMS_1_19_Common_ChatPreviewHandler`. `NMS_1_19_R1_ChatPreviewHandler` and `NMS_1_19_1_R1_ChatPreviewHandler` extend this class, and it handles things like intercepting chat preview packets and converting the chatPreviewQuery into JSON, which was previously duplicated across those two classes. Along with reorganizing some of the code, these changes saved ~2.4 kB from the plugin jar and made the code easier to understand.

In the future, I think that `NMS_1_19_Common_ChatPreviewHandler` will be able to be adapted for new versions. The only line that looks problematic is 
```java
this.connection = ((CraftPlayer) player).getHandle().connection.connection;
```
which accesses `org.bukkit.craftbukkit.v1_19_R1.entity.CraftPlayer`. Other than that, subclasses just need to override `handleChatPreviewPacket` to do version-specific field accesses and PreviewChaching or whatever they do with it in the future.

There are two additional changes included here as well:
One, a debug message will now be sent to the console stating whether or not chat preview is active. This is small but useful when messing with chat preview.

Two, the ChatPreviewHandler will check if a previewable argument is currently being filled in before it consumes the chat preview packet. This was an attempt to address this [comment on discord](https://discord.com/channels/745416925924032523/745419648970784821/1009196184377770075) and let chat preview packets that are unrelated to the CommandAPI get processed as normal.

These changes were tested on 1.19 and 1.19.2, and no problems were noticed.

### DIFF
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
@@ -230,8 +230,9 @@ public final class CommandAPI {
 	
 			}, plugin);
 			logNormal("Chat preview enabled");
-		} else
+		} else {
 			logNormal("Chat preview is not available");
+		}
 
 		CommandAPIHandler.getInstance().getPaper().registerReloadHandler(plugin);
 	}
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
@@ -229,7 +229,9 @@ public final class CommandAPI {
 				}
 	
 			}, plugin);
-		}
+			logNormal("Chat preview enabled");
+		} else
+			logNormal("Chat preview is not available");
 
 		CommandAPIHandler.getInstance().getPaper().registerReloadHandler(plugin);
 	}
--- a/commandapi-nms/commandapi-1.19-common/src/main/java/dev/jorel/commandapi/nms/NMS_1_19_Common_ChatPreviewHandler.java
+++ b/commandapi-nms/commandapi-1.19-common/src/main/java/dev/jorel/commandapi/nms/NMS_1_19_Common_ChatPreviewHandler.java
@@ -1,0 +1,122 @@
+package dev.jorel.commandapi.nms;
+
+import com.mojang.brigadier.ParseResults;
+import com.mojang.brigadier.context.ParsedCommandNode;
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
+import dev.jorel.commandapi.CommandAPIHandler;
+import dev.jorel.commandapi.arguments.PreviewInfo;
+import dev.jorel.commandapi.exceptions.WrapperCommandSyntaxException;
+import dev.jorel.commandapi.wrappers.PreviewableFunction;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandlerContext;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.serializer.gson.GsonComponentSerializer;
+import net.md_5.bungee.api.chat.BaseComponent;
+import net.md_5.bungee.api.chat.TextComponent;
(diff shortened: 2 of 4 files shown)


commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
```diff
@@ -229,7 +229,9 @@ public void onPlayerQuit(PlayerQuitEvent e) {
 				}
 	
 			}, plugin);
-		}
+			logNormal("Chat preview enabled");
+		} else
+			logNormal("Chat preview is not available");
```
JorelAli, line 234: I highly discourage the use of if/else statements without braces. The following should be used instead:

```java
} else {
    logNormal("Chat preview is not available");
}
```

commandapi-nms/commandapi-1.19-common/src/main/java/dev/jorel/commandapi/nms/NMS_1_19_Common_ChatPreviewHandler.java
```diff
@@ -0,0 +1,136 @@
...
+		ParseResults<CommandSourceStack> results = nms.getBrigadierDispatcher().parse(fullInput, nms.getCLWFromCommandSender(this.player));
+
+		// Generate the path for lookup
+		List<String> path = new ArrayList<>();
+		for (ParsedCommandNode<CommandSourceStack> commandNode : results.getContext().getNodes()) {
+			path.add(commandNode.getNode().getName());
+		}
+		Optional<PreviewableFunction<?>> preview = CommandAPIHandler.getInstance().lookupPreviewable(path);
+		if (preview.isEmpty())
+			return null;
```
JorelAli, line 86: Use of if/else without braces

commandapi-nms/commandapi-1.19-common/src/main/java/dev/jorel/commandapi/nms/NMS_1_19_Common_ChatPreviewHandler.java
```diff
@@ -0,0 +1,137 @@
...
+			}
+		}
+
+		// Normal packet handling
+		super.channelRead(ctx, msg);
+	}
+
+	protected abstract void handleChatPreviewPacket(ServerboundChatPreviewPacket chatPreview);
+
+	public MutableComponent chatPreviewQueryToJson(String chatPreviewQuery) {
```
JorelAli, line 74: This method name is slightly misleading, because the return type isn't JSON, it's an NMS component. `chatPreviewToNMSComponent` may be a more suitable name.

commandapi-nms/commandapi-1.19-common/src/main/java/dev/jorel/commandapi/nms/NMS_1_19_Common_ChatPreviewHandler.java
```diff
@@ -0,0 +1,137 @@
...
+		// Substring 1 to get rid of the leading /
+		final String fullInput = chatPreviewQuery.substring(1);
+		ParseResults<CommandSourceStack> results = nms.getBrigadierDispatcher().parse(fullInput, nms.getCLWFromCommandSender(this.player));
+
+		// Generate the path for lookup
+		List<String> path = new ArrayList<>();
+		for (ParsedCommandNode<CommandSourceStack> commandNode : results.getContext().getNodes()) {
+			path.add(commandNode.getNode().getName());
+		}
+		Optional<PreviewableFunction<?>> preview = CommandAPIHandler.getInstance().lookupPreviewable(path);
```
willkroboth, line 84: The variables for `fullInput`, `results`, `path`, and `preview` were already calculated once before in the channelRead method when checking if the preview query was worth consuming. Since the results are the same for the same chatPreviewQuery input, these variables could conceivably be stored by channelRead to be used later when parseChatPreviewQuery is called.

IIRC, Brigadier says that calling parse is an expensive operation and the results should be cached whenever possible, so maybe this is worth doing?

PR #335, opened 2022-09-04
Add CommandArgument

### LINKED ISSUE #307: Add a greedy "command argument"
### Description

It is currently possible to import Brigadier and create a GreedyStringArgument that lets users input a Minecraft command as an argument like so:

From https://commandapi.jorel.dev/8.4.0/brigadiersuggestions.html
```java
ArgumentSuggestions commandSuggestions = (info, builder) -> {
    // The current argument, which is a full command
    String arg = info.currentArg();

    // Identify the position of the current argument
    int start;
    if(arg.contains(" ")) {
        // Current argument contains spaces - it starts after the last space and after the start of this argument.
        start = builder.getStart() + arg.lastIndexOf(' ') + 1;
    } else {
        // Input starts at the start of this argument
        start = builder.getStart();
    }
    
    // Parse command using brigadier
    ParseResults<?> parseResults = Brigadier.getCommandDispatcher()
        .parse(info.currentArg(), Brigadier.getBrigadierSourceFromCommandSender(info.sender()));
    
    // Intercept any parsing errors indicating an invalid command
    for(CommandSyntaxException exception : parseResults.getExceptions().values()) {
        // Raise the error, with the cursor offset to line up with the argument
        throw new CommandSyntaxException(exception.getType(), exception.getRawMessage(), exception.getInput(), exception.getCursor() + start);
    }

    return Brigadier
        .getCommandDispatcher()
        .getCompletionSuggestions(parseResults)
        .thenApply((suggestionsObject) -> {
            // Brigadier's suggestions
            Suggestions suggestions = (Suggestions) suggestionsObject;

            return new Suggestions(
                // Offset the index range of the suggestions by the start of the current argument
                new StringRange(start, start + suggestions.getRange().getLength()),
                // Copy the suggestions
                suggestions.getList()
            );
        });
};

new CommandAPICommand("commandargument")
    .withArguments(new GreedyStringArgument("command").replaceSuggestions(commandSuggestions))
    .executes((sender, args) -> {
        // Run the command using Bukkit.dispatchCommand()
        Bukkit.dispatchCommand(sender, (String) args[0]);
    }).register();
```
Similar to how the ListArgument is a GreedyStringArgument that automatically applies a special ArgumentSuggesstions rule (#275), it would be useful if there was a CommandArgument that implemented this special ArgumentSuggestions for you. This would let users create commands like the vanilla `/execute run [another command]` without having to import Brigadier or figuring out the complicated ArgumentSuggestions.

### Expected code

The same example from before, but using the CommandArgument (much simpler)
```java
new CommandAPICommand("commandargument")
                .withArguments(new CommandArgument("command"))
                .executes((sender, args) -> {
                    // Run the command using Bukkit.dispatchCommand()
                    Bukkit.dispatchCommand(sender, (String) args[0]);
                }).register();
```

A sudo command for running a command as another player
```java
new CommandAPICommand("sudo")
                .withArguments(
                        new PlayerArgument("target"),
                        new CommandArgument("command")
                ).executes(
                        (sender, args) -> {
                            Player target = (Player) args[0];
                            String command = (String) args[1];
                            Bukkit.dispatchCommand(target, command);
                        }
                ).register();
```

### Extra details

I don't know if the Brigadier methods used already do this, but it would be nice if the CommandArgument could also receive plugin commands. Maybe this can only work after converting the plugin commands using the CommandAPI.

willkroboth:
Resolves #307 
This PR adds the CommandArgument, based on [this example](https://commandapi.jorel.dev/8.4.0/brigadiersuggestions.html#example---using-a-minecraft-command-as-an-argument) for Brigadier Suggestions in the documentation. Improving upon that example, this CommandArgument uses the server's CommandMap to provide tab-completions for all commands registered on the server, plugin and Brigadier alike.

The object returned by the CommandArgument is CommandResult. This is just a simple record that bundles the Command and the arguments together, like so:
```java
public record CommandResult(Command command, String[] args) {
	public void execute(CommandSender sender) {
		command.execute(sender, command.getLabel(), args);
	}
}
```
`CommandResult#execute` makes it easier for developers to use the CommandResult without picking apart the record if they don't need to be that specific.

The usual `replaceSuggestions(ArgumentSuggestions)` method was upgraded to  `replaceSuggestions(ArgumentSuggesstions...)` for the CommandArgument. This allows developers to override each argument in the command with their own suggestions, like so:
```java
new CommandAPICommand("restrictedcommand")
	.withArguments(new CommandArgument("command")
		.replaceSuggestions(
			ArgumentSuggestions.strings("give"),
			ArgumentSuggestions.strings(info -> Bukkit.getOnlinePlayers().stream().map(Player::getName).toArray(String[]::new)),
			ArgumentSuggestions.strings("diamond", "minecraft:diamond"),
			ArgumentSuggestions.empty()
		)
	)
```
The CommandArgument of `/restrictedcommand` will only allow commands of the form `give [online player] (diamond/minecraft:diamond)`. If a user tries to enter any other command the next suggestions will stop showing up and when the command is executed an appropriate CommandSyntaxException will be thrown. Developers can also pass in `null` instead of an ArgumentSuggestions to use the regular tab completions provided by the server.

In order to make an ArgumentTest for the CommandArgument, `NMS#getSimpleCommandMap` was implemented in MockNMS. I also added the method `public <T> void assertStoresResult(CommandSender sender, String command, Mut<T> queue, T expected)` to the ArgumentTests class to reduce code redundancy when asserting the parsed result from running a command.

Documentation for the CommandArgument still needs to be written. Additionally, the example for Brigadier suggestions should probably be updated. Maybe [this code](https://discord.com/channels/745416925924032523/745757702608912556/1013426717081227314) from discord would be a good replacement.


commandapi-core/src/main/java/dev/jorel/commandapi/arguments/CommandAPIArgumentType.java
```diff
@@ -82,6 +82,11 @@ public enum CommandAPIArgumentType {
 	 */
 	CHATCOLOR("minecraft:color"),
 
+	/**
+	 * The CommandArgument
+	 */
+	COMMAND,
```
JorelAli, line 88: This needs a string name and an implementation in `commandapi-plugin`'s `AdvancedConverter.java` file. The name `api:command` is sufficient. Adding this allows the plugin command converter to use the command argument.

I'm aware that `LIST` doesn't have one - that's simply because the list argument is much much more complex than this argument and I can see this argument being used (e.g. for a plugin converted `/sudo` command)

commandapi-core/src/main/java/dev/jorel/commandapi/arguments/CommandResult.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.Arrays;
+
+public record CommandResult(Command command, String[] args) {
+	public void execute(CommandSender sender) {
+		command.execute(sender, command.getLabel(), args);
+	}
+
+	@Override
+	public boolean equals(Object o) {
```
JorelAli, line 14: CodeFactor complains that this record doesn't have a corresponding `hashCode()` method:

> The contract of `equals()` and `hashCode()` requires that equal objects have the same hashCode. Therefore, whenever you override `equals()` you must override `hashCode()` to ensure that your class can be used in hash-based collections.

We'll need a `hashCode()` method for this record.

commandapi-core/src/main/java/dev/jorel/commandapi/arguments/CommandArgument.java
```diff
@@ -0,0 +1,203 @@
...
+		});
+	}
+
+	private boolean hasReplacement(int index) {
+		return replacements.length > index && replacements[index] != null;
+	}
+
+	ArgumentSuggestions[] replacements = new ArgumentSuggestions[0];
+
+	public CommandArgument replaceSuggestions(ArgumentSuggestions... suggestions) {
```
JorelAli, line 133: Needs JavaDocs.

JorelAli: The suggestion replacing using `replaceSuggestions(ArgumentSuggestions...)` is cool, but there is one problem with it: It doesn't support _multiple commands_.

If I (for some bizarre reason) wanted to have a command argument that accepts two commands (e.g. `/give [online player] (diamond/minecraft:diamond)` and `/tp [online player] [target player]`), you can't do that with the current implementation (as far as I'm aware).

willkroboth: Multiple commands can be supported like this:
```java
new CommandAPICommand("multiplecommands")
        .withArguments(
                new CommandArgument("command").replaceSuggestions(
                        ArgumentSuggestions.strings("give", "tp"),
                        ArgumentSuggestions.strings(info -> Bukkit.getOnlinePlayers().stream().map(Player::getName).toArray(String[]::new)),
                        ArgumentSuggestions.strings(info -> switch ((String) info.previousArgs()[0]) {
                            case "give" -> new String[]{"diamond", "minecraft:diamond"};
                            case "tp" -> Bukkit.getOnlinePlayers().stream().map(Player::getName).toArray(String[]::new);
                            // This default case should never happen thanks to CommandArgument's suggestion enforcement
                            default -> throw new IllegalStateException("Only give or tp command should have been passed in");
                        })
                )
        ).executes((sender, args) -> {
            ((CommandResult) args[0]).execute(sender);
        }).register();
```
Since the previousArgs given to the SuggestionInfo includes the command label, you can use that to switch between different suggestion paths. I suppose this is an important detail to emphasize: the SuggestionInfo object here holds information for the command being input to the CommandArgument and you cannot access the information from the arguments before the CommandArgument.

Perhaps a more intuitive system could be developed with some sort of tree structure that leads to different ArgumentSuggestions objects, but I think this makes enough sense.

JorelAli: Good point - I didn't think of that. It's a bit clunky (which can be mitigated by having suitable documentation which isn't a big issue), but I think having a cleaner implementation that's easier for users may be worth considering. Perhaps something of the form...

```java
replaceSuggestions(SomeClassName...)
```

```java
.replaceSuggestions(
	SomeClassName.of(
		ArgumentSuggestions.strings("give"),
		ArgumentSuggestions.strings(info -> Bukkit.getOnlinePlayers().stream().map(Player::getName).toArray(String[]::new)),
		ArgumentSuggestions.strings("diamond", "minecraft:diamond"),
		ArgumentSuggestions.empty()
	),
	SomeClassName.of(
		ArgumentSuggestions.strings("tp"),
		ArgumentSuggestions.strings(info -> Bukkit.getOnlinePlayers().stream().map(Player::getName).toArray(String[]::new)),
		ArgumentSuggestions.strings(info -> Bukkit.getOnlinePlayers().stream().map(Player::getName).toArray(String[]::new)),
		ArgumentSuggestions.empty()
	)
)
```

It's literally syntactic sugar, but I think it'll make this feature more discoverable and save on dev brain power. This could be a possibility _in addition to_ the existing `replaceSuggestions(ArgumentSuggestions... suggestions)` - I think the method you've already implemented makes a lot of sense.

JorelAli: > Needs JavaDocs.

Can confirm implemented.

willkroboth: Yeah, another method in CommandArgument like `replaceBranchingSuggesstions` might be nice. I'll try something and see how it goes.

willkroboth:
Multiple paths for command suggestions are now possible with this syntax (taken from ArgumentTests):
```java
new CommandArgument("command")
	.branchSuggestions(
		SuggestionsBranch.suggest(
			ArgumentSuggestions.strings("give"),
			ArgumentSuggestions.strings(info -> Bukkit.getOnlinePlayers().stream().map(Player::getName).toArray(String[]::new))
		).branch(
			SuggestionsBranch.suggest(
				ArgumentSuggestions.strings("diamond", "minecraft:diamond"),
				ArgumentSuggestions.empty()
			),
			SuggestionsBranch.suggest(
				ArgumentSuggestions.strings("dirt", "minecraft:dirt"),
				null,
				ArgumentSuggestions.empty()
			)
		),
		SuggestionsBranch.suggest(
			ArgumentSuggestions.strings("tp"),
			ArgumentSuggestions.strings(info -> Bukkit.getOnlinePlayers().stream().map(Player::getName).toArray(String[]::new)),
			ArgumentSuggestions.strings(info -> Bukkit.getOnlinePlayers().stream().map(Player::getName).toArray(String[]::new))
		)
	)
```
`CommandArgument#replaceSuggestions` and `SuggestionsBranch#suggest` add ArgumentSuggestions, then `CommandArgument#branchSuggestions` and `SuggestionsBranch#branch` are used to create new paths which themselves can be filled with ArgumentSuggestions then a branch.

PR #336, opened 2022-09-08
Allow suggestions and tooltips to be provided as collections in addition to arrays

469512345:
This PR simply adds overrides for passing collections for arguments and tooltips as well as using arrays.
The aim of this is to avoid conversion into arrays when generating suggestions or tooltips. This can be particularly irritating when generics get involved, so instead collections avoid this problem.

For example in `ArgumentSuggestions`
```java
	 /**
	 * Suggest hardcoded strings
	 *
	 * @param suggestions array of hardcoded strings
	 *
	 * @return an {@link ArgumentSuggestions} object suggesting hardcoded strings
	 */
	static ArgumentSuggestions strings(String... suggestions) {
		return (info, builder) -> future(suggestionsFromStrings(builder, suggestions));
	}
```

gets an additional overload

```java 
	/**
	 * Suggest hardcoded strings
	 *
	 * @param suggestions collection of hardcoded strings
	 *
	 * @return an {@link ArgumentSuggestions} object suggesting hardcoded strings
	 */
	static ArgumentSuggestions strings(Collection<String> suggestions) {
		return (info, builder) -> future(suggestionsFromStrings(builder, suggestions));
	}
```
This is repeated for all methods in `ArgumentSuggestions`, `SafeSuggestions`, `Tooltip`, and `StringTooltip` that use array arguments or varargs.

Unfortunately due to Generic erasure with the `Function` type, it was not possible to keep all overloads with the same name

```java
	/**
	 * Suggest an array of strings as the result of a function
	 *
	 * @param suggestions function providing the strings as an array
	 *
	 * @return an {@link ArgumentSuggestions} object suggesting the result of the function
	 */
	static ArgumentSuggestions strings(Function<SuggestionInfo, String[]> suggestions) {
		return (info, builder) -> future(suggestionsFromStrings(builder, suggestions.apply(info)));
	}

	/**
	 * Suggest a collection of strings as the result of a function
	 *
	 * @param suggestions function providing the strings as a collection
	 *
	 * @return an {@link ArgumentSuggestions} object suggesting the result of the function
	 */
	static ArgumentSuggestions stringCollection(Function<SuggestionInfo, Collection<String>> suggestions) {
		return (info, builder) -> future(suggestionsFromStrings(builder, suggestions.apply(info)));
	}
```

In my opinion it's not worth deprecating the original method in favour of something like `stringArray(Function<SuggestionInfo, String[]>)` but this is of course possible if required.

The return type of methods in `Tooltip` and `StringTooltip` have all been migrated to use Collection instead of array to avoid problems with generic array creation and unchecked casts.

It doesn't entirely sit right with me just how many methods now exist in these classes, but this is due to the inflexibility of Java's type system. This is also why I created `ArgumentSuggestions` and `SafeSuggestions` classes, to store these factory methods without needing to modify the `Argument` implementation.

### DIFF
--- a/commandapi-core/pom.xml
+++ b/commandapi-core/pom.xml
@@ -46,7 +46,7 @@
 			<groupId>net.kyori</groupId>
 			<artifactId>adventure-platform-bukkit</artifactId>
 			<version>4.1.0</version>
-			<scope>provided</scope>
+			<scope>test</scope>
 		</dependency>
 		<dependency>
 			<groupId>io.papermc.paper</groupId>
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/StringTooltip.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/StringTooltip.java
@@ -24,8 +24,11 @@ import com.mojang.brigadier.Message;
 import net.kyori.adventure.text.Component;
 import net.md_5.bungee.api.chat.BaseComponent;
 
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.stream.Stream;
 
 /**
  * Represents a suggestion for an argument with a hover tooltip text for that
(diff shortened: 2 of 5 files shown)


JorelAli:
Sounds good to me 👍
I'm aware there are lots of overloads and it's a pain because it adds more bloat than actually necessary, but these tiny details makes the API easier to use and that's always a big plus.

PR #357, opened 2022-10-27
adding Kotlin DSL

DerEchtePilz:
This adds a Kotlin DSL.

If you want me to change anything, I'll do it - especially in the documentation because I used my English I know (as I am not a native speaker).


commandapi-kotlin/src/test/kotlin/Examples.kt
```diff
@@ -0,0 +1,87 @@
...
+    playerArgument("player") {
+        greedyArgument("msg") {
+            playerExecutor { _, args ->
+                val player: Player = args[0] as Player
+                val message: String = args[1] as String
+                player.sendMessage(message)
+            }
+        }
+    }
+    requirement(of("broadcast"), { sender: CommandSender -> (sender is Player) && sender.isOp }) { // Define a new LiteralArgument("broadcast") that requires the CommandSender to be a player who is a server operator
```
JorelAli, line 36: The check for whether the sender is a player is redundant, because this command only uses `playerExecutor`. The CommandAPI documentation [already has a similar case](https://commandapi.jorel.dev/8.5.1/tooltips.html#example---teleportation-command-with-suggestion-descriptions) where we already know the cast will work as long as we use `.executesPlayer`.

docssrc/src/kotlinintro.md
```diff
@@ -0,0 +1,92 @@
+# Kotlin-based commands
+
+The CommandAPI also provides an alternative way of making commands when using Kotlin to develop your plugins: A DSL!
+
+This DSL provides many methods to easily add arguments to your command structure. Examples of the DSL can be found [here](./kotlinusedsl.md)
```
JorelAli, line 5: Sentence is missing a full stop

docssrc/src/kotlinusedsl.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Executors
+
+The CommandTree DSL also provides executors to execute your command. You've seen the `anyExecutor` in the example above.
+
+To find out, which DSL executor corresponds to "normal" executors, you can refer to the table below:
+
+| DSL Executor           | "normal" Executor      |
+|------------------------|------------------------|
+| anyExecutor()          | executes()             |
```
JorelAli, line 29: Code should be formatted as inline code in a table (surrounded with ` symbols)

docssrc/src/kotlinusedsl.md
```diff
@@ -0,0 +1,127 @@
...
+
+The way arguments are implemented is pretty straight forward: It's basically the argument class' name, but as a method. So if you wanted to use a `ItemStackArgument` in your command, you would use the `itemStackArgument()` method of the DSL.
+
+One thing to note is that the DSL also features every existing constructor. This means if you want to use an `IntegerArgument` with a minimum of `0` and a maximum of `10`, you normally would implement it like this: 
+
+`new IntegerArgument("integer", 0, 10)`
+
+However, when using this DSL it is implemented like this: 
+
+`integerArgument("integer", 0, 10) { // new arguments, execution goes in here }`
```
JorelAli, line 61: These examples should be a full code block, not inline code.

The following is invalid Kotlin - you should use `/*` and `*/` for the comment inside, instead of `//`
```kt
integerArgument("integer", 0, 10) { // new arguments, execution goes in here }
```

commandapi-kotlin/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+			<groupId>io.papermc.paper</groupId>
+			<artifactId>paper-api</artifactId>
+			<version>${paper.version}</version>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.jetbrains.kotlin</groupId>
+			<artifactId>kotlin-stdlib</artifactId>
+			<version>${kotlin.version}</version>
+		</dependency>
```
JorelAli, line 117: This is leaking the Kotlin dependency to end users - the end developer can access this Kotlin library without declaring it in their `pom.xml`. This can lead to shading conflicts - if the end developer declares Kotlin themselves, it may conflict with this kotlin here. (You should use `<scope>provided</scope>`)

commandapi-kotlin/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+			<groupId>org.jetbrains.kotlin</groupId>
+			<artifactId>kotlin-test</artifactId>
+			<version>${kotlin.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>dev.jorel</groupId>
+			<artifactId>commandapi-core</artifactId>
+			<version>${project.version}</version>
+		</dependency>
```
JorelAli, line 128: If you don't add `<scope>provided</scope>` here, this also gets shaded into `commandapi-kotlin` when it shouldn't!

commandapi-kotlin/pom.xml
DerEchtePilz: Should I then add the instructions of setting up Kotlin to the documentation?

commandapi-kotlin/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+						<id>test-compile</id>
+						<phase>test-compile</phase>
+						<goals>
+							<goal>test-compile</goal>
+						</goals>
+					</execution>
+				</executions>
+				<configuration>
+					<jvmTarget>1.8</jvmTarget>
+				</configuration>
```
JorelAli, line 59: What? The CommandAPI uses Java 16

commandapi-kotlin/pom.xml
JorelAli: Yes please.

PR #362, opened 2022-11-11
fix for overloaded methods being placed next to each other.

AkaGiant:
This is mostly just a test to see if CodeFactor is crying about it. 

I think you want to decline this more on the fact that it does not fix anything.


AkaGiant:
I was expecting code factor to find +1 fix.

Issue #358, opened 2022-10-31
Add miscellaneous argument type 'World'


lokka30:
### Description

Allows us to select world names, such as:

- `/kill @e[type=creeper] my_cool_world`

..where `my_cool_world` is a world loaded on the server.

Can also be used as a list:

- `/kill @e[type=creeper] my_cool_world world world_nether world_the_end resource_world`

lokka30:
Unfortunately the current workaround doesn't seem to work on `load: STARTUP` plugins, nor on worlds which have been loaded after the plugin's commands were registered.
I have tried a workaround where I have replaced the suggestions with a list of worlds which are queried async, of course this hasn't resolved the issue since the list of allowed list values have not been modified.
For now, I will make it use a string list argument and parse worlds manually. 🙂

lokka30:
Thanks, no worries :)

JorelAli:
After a quick chat with a member from the CommandAPI Discord server, I've discovered that the CommandAPI's implementation of Minecraft's `DimensionArgument` is terrible and doesn't do what it is supposed to!

Scheduling the implementation of a `DimensionArgument` for CommandAPI 8.6.0 which will do pretty much exactly what you would expect - it returns a Bukkit `World` object.

JorelAli:
Implemented in release [8.6.0](https://github.com/JorelAli/CommandAPI/releases/tag/8.6.0).

PR #365, opened 2022-11-19
adding WorldArgument

DerEchtePilz:
This adds the WorldArgument suggested in #358 
Feel free to merge it (or postpone it for a later version), I just wanted to implement this because it sounded simple enough for me xD

### DIFF
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/arguments/WorldArgument.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/arguments/WorldArgument.java
@@ -28,21 +28,11 @@ public class WorldArgument extends Argument<World> {
 		super.replaceSuggestions((info, builder) -> {
 			String currentArg = info.currentArg();
 
-			String[] worlds = new String[Bukkit.getWorlds().size()];
-			List<World> worldsList = Bukkit.getWorlds();
-			for (World world : worldsList) {
-				worlds[worldsList.indexOf(world)] = world.getName();
-			}
-
-			List<String> worldSuggestions = new ArrayList<>();
-			for (String world : worlds) {
-				if (world.startsWith(currentArg)) {
-					worldSuggestions.add(world);
+			for (World world : Bukkit.getWorlds()) {
+				String worldName = world.getName();
+				if (worldName.startsWith(currentArg)) {
+					builder.suggest(worldName);
 				}
-			}
-
-			for (String result : worldSuggestions) {
-				builder.suggest(result);
 			}
 
 			return builder.buildFuture();
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/arguments/WorldArgument.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/arguments/WorldArgument.java
@@ -1,0 +1,75 @@
+package dev.jorel.commandapi.arguments;
+
+import com.mojang.brigadier.StringReader;
+import com.mojang.brigadier.arguments.StringArgumentType;
+import com.mojang.brigadier.context.CommandContext;
+import com.mojang.brigadier.exceptions.CommandSyntaxException;
+import dev.jorel.commandapi.nms.NMS;
+import org.bukkit.Bukkit;
+import org.bukkit.World;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An argument that represents the Bukkit {@link World} object
+ *
+ * @apiNote Returns a {@link World} object
+ */
+public class WorldArgument extends Argument<World> {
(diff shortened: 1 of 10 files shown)


DerEchtePilz:
Oops, forgot to add it to the Kotlin DSL

commandapi-core/src/test/kotlin/Examples.kt
```diff
@@ -1835,7 +1755,7 @@ getServer().addRecipe(emeraldSwordRecipe)
 // Safely override with the recipe we've defined
 val arguments = listOf<Argument<*>>(
     RecipeArgument("recipe").replaceSafeSuggestions(SafeSuggestions.suggest {
-        arrayOf(emeraldSwordRecipe, /* Other recipes here */)
+        arrayOf(emeraldSwordRecipe /* Other recipes here */)
```
JorelAli, line 1758: This comma was intentional. It shouldn't have been removed.

DerEchtePilz: I didn't even know I removed that xD Will add that back now

commandapi-core/src/main/java/dev/jorel/commandapi/arguments/WorldArgument.java
```diff
@@ -0,0 +1,75 @@
...
+			List<String> worldSuggestions = new ArrayList<>();
+			for (String world : worlds) {
+				if (world.startsWith(currentArg)) {
+					worldSuggestions.add(world);
+				}
+			}
+
+			for (String result : worldSuggestions) {
+				builder.suggest(result);
+			}
```
JorelAli, line 46: This looks like a _really really really_ complicated way of doing:
```java
for (World world : Bukkit.getWorlds()) {
	final String worldName = world.getName();
	if(worldName.startsWith(currentArg)) {
		builder.suggest(worldName);
	}
}
```
Care to justify why you implemented this with three for loops?

willkroboth: Also, I don't think you need to check if `worldName.startsWith(currentArg)`, because I think `builder#suggest` already does that (relating to [this comment in discord](https://discord.com/channels/745416925924032523/745418840153784401/1042923222393421904)).

DerEchtePilz: Yeah... no. After seeing your solution, I actually feel pretty dumb when looking at my solution...

DerEchtePilz: > Also, I don't think you need to check if `worldName.startsWith(currentArg)`, because I think `builder#suggest` already does that (relating to [this comment in discord](https://discord.com/channels/745416925924032523/745418840153784401/1042923222393421904)).

Actually, I tried it without that first and the suggestions didn't work at all.

willkroboth: Oh, huh, looking at `builder#suggest`, here's what it has:
```java
 public SuggestionsBuilder suggest(String text) {
    if (text.equals(this.remaining)) {
        return this;
    } else {
        this.result.add(new Suggestion(StringRange.between(this.start, this.input.length()), text));
        return this;
    }
}
```
I think I was confusing that with [`ArgumentSuggestions#shouldSuggest`](https://github.com/JorelAli/CommandAPI/blob/dev/dev/commandapi-core/src/main/java/dev/jorel/commandapi/arguments/ArgumentSuggestions.java#L282) which dose filter out suggestions with `.startsWith`.

JorelAli:
I appreciate the work put into this PR, but I'm closing this. After a long think about things, 
the CommandAPI will only be accepting "custom argument implementations" that satisfy the following points:

- The custom argument is an implementation of a Vanilla Minecraft argument type that is specified in [Argument types](https://minecraft.fandom.com/wiki/Argument_types).
- The custom argument implementation heavily requires the use of Brigadier or NMS.

Hopefully, these two points above cover all bases, including new upcoming arguments that are added in future Minecraft updates, as well as complicated arguments such as the newly added `CommandArgument`.

The CommandAPI documentation already includes an example of a world argument [here](https://commandapi.jorel.dev/8.5.1/customarguments.html?highlight=World,argumen#example---world-argument) which makes use of the `CustomArgument` feature. This implementation using a `CustomArgument` does not require any additional libraries (e.g. Brigadier), nor does it require code which is hard/impractical to access (e.g. NMS). This makes it a perfect use case for end-user developers to implement themselves, rather than adding to the core CommandAPI.

Issue #360, opened 2022-11-11
CommandAPI command executors should have an overload for arguments accessible via a map


JorelAli:
### Description

CommandAPI command executors currently only consist of a command sender and an array of arguments:

```java
(sender, args) -> {
    // ...
}
```

It is useful to be able to access arguments by node name, rather than by index - especially in the case of using optional arguments:

```java
(CommandSender sender, Map<String, Object> args) -> {
    String input = (String) args.get("blah");
    int amount = (int) args.getOrDefault("amount", 1);
}
```

It would be ideal to implement an overload for the various `executes()` methods that instead of accepting a `(sender, args)`, it can instead accept a generalized `(executioninfo)` which would allow the CommandAPI to be more easily extensible in the future and provide users with a more accessible API.

### Expected code

This could be implemented using a `record` instead:
```java
class ExecutionInfo<Sender extends CommandSender> {
  public Object[] args();
  public Sender sender();
  public Map<String, Object> argsMap();
}
```

Generic structure for a player execution, using the generic type to enforce the return type of `sender()` as as `Player` object:
```java
@FunctionalInterface
public interface PlayerExecution {

    public void run(ExecutionInfo<Player> info);

}
```

The implementation of `argsMap` would be as simple as tweaking the implementation of `CommandAPIHandler.argsToObjectArr()`, and shoving them into a `LinkedHashMap`. We'll use `Map` as the main interface for `argsMap()` for simplicity, but mention in JavaDocs that the implementation is declared in order (so can be iterated upon if users so desire...)

willkroboth:
For those who want extra reading on the context of this suggestion, it started as a [comment in discord](https://discord.com/channels/745416925924032523/745757702608912556/1040618915476287509)

Sculas:
My idea was: (as discussed on [Discord](https://discord.com/channels/745416925924032523/745418987948736543/1045720890178011207))
```java
new CommandAPICommand("suggestion")
    .withArguments(new ItemStackArgument("item"))
    .executesPlayer(info -> {
        ItemStack possibilityOne = info.args().get(0); // @Nullable public <T> T get(int index)
        ItemStack possibilityTwo = info.args().get("item"); // @Nullable public <T> T get(String identifier)
        // ^ @Nullable because the developer should handle these cases, it shouldn't throw a "hidden" exception

        info.player().getInventory().addItem(possibilityOne, possibilityTwo);
    })
    .register();
```
[...]

PR #369, opened 2022-11-24
adding #360

DerEchtePilz:
Hello!

It's me again.

I opened this PR so you can decide if you want to have it in 8.6.0 or later (or if you want me to change/add anything).


DerEchtePilz:
Also, just wanted to mention that this currently is untested which will be done until noon on Saturday.

DerEchtePilz:
As requested on [discord](https://discord.com/channels/745416925924032523/745419648970784821/1045458314605113384), I will now add some code examples and explain my changes a little more.

### 1. Internal changes
To be able to implement this without having to use if-statements (in fact, I didn't use a single one), I changed the `executeWith` method in the `IExecutorTyped` interface to now also take in a `Map<String, Object>`. This map is filled with content in `CommandAPIHandler#generateCommand`.

#### 1.1 Interfaces and `executes...`
Before I made this PR, each executor had two interfaces: `...CommandExecutor` and `...ResultingCommandExecutor`.
Now, two more interfaces have been added to support the use of the map:
- `...ExecutionInfo`
- `...ResultingExecutionInfo`

The `...ExecutionInfo` and `...ResultingExecutionInfo` interfaces both have one method:
- `...ExecutionInfo`:
```java
void run(BlockCommandSender sender, Object[] args, Map<String, Object> argsMap) throws WrapperCommandSyntaxException;
```
- `...ResultingExecutionInfo`:
```java
int run(BlockCommandSender sender, Object[] args, Map<String, Object> argsMap) throws WrapperCommandSyntaxException;
```

Those two interfaces are now also available through the use of the `.executes...` methods.
The already existing `.executes...` methods have now been overloaded to also accept the new ExecutionInfo interfaces:
```java
public T executesPlayer(PlayerExecutionInfo info) {
	this.executor.addNormalExecutor(info);
	return (T) this;
}

public T executesPlayer(PlayerResultingExecutionInfo info) {
	this.executor.addResultingExecutor(info);
	return (T) this;
}
```

#### Use the new methods in your code
For this section, I will just add some examples I wrote:
```java
new CommandAPICommand("nomap")
	.withArguments(new ItemStackArgument("item"))
	.executesPlayer((player, args) -> {
		// This command does not have access to the map with stores the arguments mapped to their node names
		// We have to get the ItemStack as normal
		ItemStack itemStack = (ItemStack) args[0];
		player.getInventory().addItem(itemStack);
	})
	.register();

new CommandAPICommand("withmap")
	.withArguments(new ItemStackArgument("item"))
	.executesPlayer((player, args, argsMap) -> {
		// This command has access to the map which stores the arguments mapped to their node names
		// Here we can get the ItemStack with two ways:
		// The known way:
		ItemStack possibilityOne = (ItemStack) args[0];

		// Or:
		ItemStack possibilityTwo = (ItemStack) argsMap.get("item");

		player.getInventory().addItem(possibilityOne, possibilityTwo);
	})
	.register();
```

### 2. Tests
As written previously, I did not test this yet but I am confident that it should work.
I will do that until Saturday und will update you when I am done.

JorelAli:
Not scheduling for 8.6.0 - too close to the expected release date. I'm more than happy to have this for 9.0.0.

My problem with this implementation is the limited scope for future extensibility. The CommandAPI has been around for just over four years and we're only now adding more accessibly functionality to the `.executes()` methods - it's quite possible that another four years down the line we may want to add more functionality to our `.executes()` method. The choice to use another functional interface with three parameters is limiting and leads into questionable design decisions, such as:
- What if we want to add a fourth parameter later? What about a fifth? We'll end up with code that gets unnecessarily large:

  ```java
  .executesPlayer((player, args, argsMap, somethingElse, someOtherStuff) -> {
  
  })
  ```

- How do we make linters happy? In programming (in general), linters like to complain about unused parameters. Ideally, the CommandAPI should avoid exposing parameters that users are unlikely to used, which is not the case with your implementation:

  ```java
  .executesPlayer((player, args, argsMap) -> {
    // Almost all users that are going to use this method will just use 'player'
    // and argsMap. We end up forcing the user to add this unused 'args' parameter.
    // In Kotlin, we can dismiss this with an underscore, but you can't do that in Java!
  })
  ```

I think going with a record would be better. It would allow us (CommandAPI development team) to be able to expand the list of accessible functions easily without causing backwards-incompatible changes and without CommandAPI users needing to change their executor implementation. Perhaps something like this would be better?

```java
new CommandAPICommand("withmap")
    .withArguments(new ItemStackArgument("item"))
    .executesPlayer(info -> {
        ItemStack possibilityOne = info.args()[0];             // Access using args()
        ItemStack possibilityTwo = info.argsMap().get("item"); // Access using argsMap()
        Player player = info.sender(); // Access command sender. Using generics, we can make this a Player instead of a CommandSender object.
    })
    .register();
```

DerEchtePilz:
Yes, I read that record suggestion in #360 and also tried to implement it like that.
I didn‘t manage to do that and decided to go with that solution, knowing it is not ideal.
I am more than happy to add that record, how would I do it in general? Would a Function work?

JorelAli:
> I am more than happy to add that record, how would I do it in general? Would a Function work?

I envision something along the lines of creating a generalized record like this:

```java
public record ExecutionInfo<Sender extends CommandSender>(
    Sender sender,
    Object[] args,
    Map<String, Object> argsMap
) {}
```

I think in the long run it would be safer to create a generic functional interface for our two types rather than using `Function<ExecutionInfo, Integer>` and `Consumer<ExecutionInfo>` (especially for Kotlin purposes). Say we add an `INormalExecution` and an `IResultingExecution` which follows something like this:

```java
public interface INormalExecution<Sender extends CommandSender> extends IExecutorTyped {

    @Override
    default int executeWith(CommandSender sender, Object[] args) throws WrapperCommandSyntaxException {
        // Some implementation here... either changing the parameters to this function
        // to using ExecutionInfo or something? I didn't think this far ahead
        return 1;
    }

    void run(ExecutionInfo<Sender> info) throws WrapperCommandSyntaxException;

}
```

Then I think it's safe to say we can add our implementations for each execution type and then link them into `Executable`. For example, for a normal player execution:

```java
@FunctionalInterface
public interface PlayerCommandExecution extends INormalExecution<Player> {

    void run(ExecutionInfo<Player> player) throws WrapperCommandSyntaxException;

    @Override
    default ExecutorType getType() {
        return ExecutorType.PLAYER;
    }
}
```

And then linking:

```java
// class dev.jorel.commandapi.Executable {

    public T executesPlayer(PlayerCommandExecution executor) {
        this.executor.addNormalExecutor(executor); // Something to this effect
        return (T) this;
     }
     
// }
```

If all of the names of "executor" and "execution" and "executable" start getting too complicated, feel free to choose alternatives for the new classes!

In short, the main idea is that we have one centralized record which is easy to update with new changes, and then each new executor implements that.

DerEchtePilz:
This is now tested and should work!

Also, I decided to not include the `CommandArgs` class which Lucas suggested on [discord](https://discord.com/channels/745416925924032523/745418987948736543/1045720890178011207) as I decided to include the `Object[]` and `Map<String, Object>` anyway.

JorelAli:
@DerEchtePilz Please update this branch with the latest changes from `JorelAli:dev/dev` which includes the multi-platform refactor so this can be merged cleanly. For all intents and purposes:
- Changes in `commandapi-core` can be moved to `commandapi-bukkit-core`
- Changes in `commandapi-kotlin` can be moved to `commandapi-kotlin-bukkit`
- Changes in `commandapi-plugin-test` can be moved to `commandapi-bukkit-plugin-test` (don't worry if this doesn't compile, this whole project is in development)

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -358,30 +358,35 @@ Command<CommandSourceStack> generateCommand(Argument<?>[] args,
...
 		List<Object> argList = new ArrayList<>();
 
+		// LinkedHashMap for arguments for executor
+		Map<String, Object> argsMap = new LinkedHashMap<>();
+
 		// Populate array
 		for (Argument<?> argument : args) {
 			if (argument.isListed()) {
 				argList.add(parseArgument(cmdCtx, argument.getNodeName(), argument, argList.toArray()));
+				argsMap.put(argument.getNodeName(), parseArgument(cmdCtx, argument.getNodeName(), argument, argList.toArray()));
```
JorelAli, line 373: We _really really really_ want to avoid parsing the argument twice. Argument parsing is fairly intense, please parse the argument only once and pass the result to the list and map objects.

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -358,30 +358,35 @@ Command<CommandSourceStack> generateCommand(Argument<?>[] args,
...
 			}
 		}
 
-		return argList.toArray();
+		Map<Integer, Object> arguments = new HashMap<>();
```
JorelAli, line 377: This looks sus. What's going on here? By the looks of things, it looks like a really weird way to implement "returning two objects at the same time". Surely here we're at a point we can just return a new `CommandArguments` object?

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -0,0 +1,26 @@
+package dev.jorel.commandapi.executors;
+
+import java.util.Map;
+
+public class CommandArguments {
```
JorelAli, line 5: This class, and all of its methods and constructors, need JavaDocs.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/ExecutionInfo.java
```diff
@@ -0,0 +1,6 @@
+package dev.jorel.commandapi.executors;
+
+import org.bukkit.command.CommandSender;
+
+public record ExecutionInfo<Sender extends CommandSender>(Sender sender, CommandArguments args) {
```
JorelAli, line 5: This class, as well as its parameters (because it's a record) need JavaDocs so developers know what these are.

JorelAli (changes requested):

DerEchtePilz:
So, what I've decided to do is that I will first resolve those issues right here (because honestly, I kinda do not want to mess anything up) and then I will update this branch.

_Can someone please tell me how to update this branch? I've merged conflicting files before but it is important (for me, you and this pull request) that everything goes well._

JorelAli (approved):

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -161,21 +163,44 @@ Command<Source> generateCommand(Argument[] args, CommandAPIExecutor<CommandSende
...
 				String[] result = new String[argsAndCmd.length - 1];
+				AbstractExecutionInfo<CommandSender, AbstractCommandSender<? extends CommandSender>> convertedExecutionInfo = new AbstractExecutionInfo<CommandSender, AbstractCommandSender<? extends CommandSender>>() {
+					@Override
+					public CommandSender sender() {
+						return sender.getSource();
+					}
+
+					@Override
+					public CommandArguments args() {
+						return new CommandArguments(result, new LinkedHashMap<>());
```
DerEchtePilz, line 192: Wait... This wouldn't work, would it?

DerEchtePilz: For more context, please take a look at the file. I don't know how to specify multiple lines (if that is possible at all)

JorelAli: Yes, this should work. Converted commands don't (and can't) access arguments by node name, they only access them via index

DerEchtePilz: Hmm... Okay, thanks. I thought it wouldn't as I pass the new String[] into that method which hasn't been filled with content yet...

commandapi-core/src/main/java/dev/jorel/commandapi/executors/AbstractExecutionInfo.java
```diff
@@ -0,0 +1,27 @@
...
+
+	/**
+	 * @return The sender of this command
+	 */
+	Sender sender();
+
+	/**
+	 * @return The wrapper type of this command
+	 */
+	WrapperType senderWrapper();
```
JorelAli, line 20: This is exposed to the end user. A comment stating that this is only used internally, and you should be using `sender()` instead would be nice.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -0,0 +1,50 @@
...
+		return args[index];
+	}
+
+	/**
+	 * Returns an argument by its node name
+	 *
+	 * @param nodeName The node name of this argument. This was set when initializing an argument
+	 * @return an argument which has the given node name
+	 */
+	public Object get(String nodeName) {
```
JorelAli, line 47: This method can return null (if `nodeName` is not found). This should be made known to the end user (by using the `javax.annotation.Nullable` annotation) and it should be stated in the JavaDocs that this can return null if `nodeName` is not found.

JorelAli (commented):
A few minor tid-bits that I noticed during manual testing, but nothing severe enough to prevent merging it over. These minor points should be addressed before 9.0.0 goes out though.

PR #388, opened 2022-12-27
Allow platform-specific configuration, Add hook-paper-reload config option

willkroboth:
The inspiration for this PR started in discord ([see here](https://discord.com/channels/745416925924032523/1056674414231552071/1056674414231552071)).

This PR does two things: 
- allows platform-specific configuration options
...which is necessary for...
- hook-paper-reload config option

First, `CommandAPIConfig` is now abstract and implements `ChainableBuilder`. Each platform should have a subclass of `CommandAPIConfig`, for example, `CommandAPIBukkitConfig`. Any information needed for configuring the platform specifically can be added to the platform implementation, such as `shouldHookPaperReload`.
```java
public abstract class CommandAPIConfig<Impl extends CommandAPIConfig<Impl>> implements ChainableBuilder<Impl> {
	// The default configuration. This should mirror the commandapi-plugin config.yml file.
	boolean verboseOutput = false;
        ...

	public Impl verboseOutput(boolean value) {
		this.verboseOutput = value;
		return instance();
	}

        ...
}

public class CommandAPIBukkitConfig extends CommandAPIConfig<CommandAPIBukkitConfig> {
	JavaPlugin plugin;
	// Default configuration
	boolean shouldHookPaperReload = true;

	public CommandAPIBukkitConfig(JavaPlugin plugin) {
		this.plugin = plugin;
	}

	public CommandAPIBukkitConfig shouldHookPaperReload(boolean hooked) {
		this.shouldHookPaperReload = hooked;

		return this;
	}

	@Override
	public CommandAPIBukkitConfig instance() {
		return this;
	}
}
```
When a platform is loaded, it will now be passed the config object, which is packaged into an appropriate subclass of `InternalConfig`, which can be accessed later.
```java
public abstract class CommandAPIBukkit<Source> ... {
	private static InternalBukkitConfig config;

	public static InternalBukkitConfig getConfiguration() {
		return config;
	}

	@Override
	public void onLoad(CommandAPIConfig<?> config) {
		if(config instanceof CommandAPIBukkitConfig bukkitConfig) {
			CommandAPIBukkit.config = new InternalBukkitConfig(bukkitConfig);
		} else {
			CommandAPI.logError("CommandAPIBukkit was loaded with non-Bukkit config!");
			CommandAPI.logError("Attempts to access Bukkit-specific config variables will fail!");
		}

		checkDependencies();
	}

	@Override
	public void onEnable() {
		JavaPlugin plugin = config.getPlugin();
		...
	}

        ...
}

public class InternalBukkitConfig extends InternalConfig {
	private final JavaPlugin plugin;
	private final boolean shouldHookPaperReload;

	public InternalBukkitConfig(CommandAPIBukkitConfig config) {
		super(config);
		this.plugin = config.plugin;
		this.shouldHookPaperReload = config.shouldHookPaperReload;
	}

	public JavaPlugin getPlugin() {
		return plugin;
	}

	public boolean shouldHookPaperReload() {
		return shouldHookPaperReload;
	}
}
```
Moving to the second point, when registering the event handler on the `ServerResourcesReloadedEvent`, the Bukkit-specific, `hook-paper-reload` config option is now considered.
```java
public class PaperImplementations {
	public void registerReloadHandler(Plugin plugin) {
		if (isPaperPresent && CommandAPIBukkit.getConfiguration().shouldHookPaperReload()) {
			Bukkit.getServer().getPluginManager().registerEvents(new Listener() {

				@EventHandler
				public void onServerReloadResources(ServerResourcesReloadedEvent event) {
					CommandAPI.logNormal("/minecraft:reload detected. Reloading CommandAPI commands!");
					nmsInstance.reloadDataPacks();
				}

			}, plugin);
			CommandAPI.logNormal("Hooked into Paper ServerResourcesReloadedEvent");
		} else {
			CommandAPI.logNormal("Did not hook into Paper ServerResourcesReloadedEvent");
		}
	}
}
```

### IMPORTANT: `CommandAPI#onLoad` and `CommandAPI#onEnable` method usage has changed
Previously, `CommandAPI#onEnable` received an `Object` that represented the 'plugin' responsible for loading it. This was done in an attempt to be as backward-compatible as possible. `CommandAPI#onEnable` used to take `JavaPlugin`, but that class is not accessible in a platform-free context, so `Object` was the best bet for that to work.

Now, however, the JavaPlugin variable was moved to the `CommandAPIBukkitConfig`. It was put in the constructor to make it required to create a Bukkit config object, reflecting how it is required for the CommandAPI to work.

Altogether, here's how shaders need to change their code (taken from maven-shaded example)
From:
```java
public class Main extends JavaPlugin {
	@Override
	public void onLoad() {
		CommandAPI.setLogger(new CommandAPIJavaLogger(getLogger()));

		CommandAPI.onLoad(
			new CommandAPIConfig()
				.verboseOutput(true)
				.dispatcherFile(new File(getDataFolder(), "command_registration.json"))
				.initializeNBTAPI(NBTContainer.class, NBTContainer::new)
		);
	}

	@Override
	public void onEnable() {
		CommandAPI.onEnable(this);

		// Register commands
		...
	}
}
```
To:
```java
public class Main extends JavaPlugin {
	@Override
	public void onLoad() {
		CommandAPI.setLogger(new CommandAPIJavaLogger(getLogger()));

		CommandAPI.onLoad(
			new CommandAPIBukkitConfig(this) // Creating a CommandAPIBukkitConfig now
				.verboseOutput(true)
				.dispatcherFile(new File(getDataFolder(), "command_registration.json"))
				.initializeNBTAPI(NBTContainer.class, NBTContainer::new)
		);
	}

	@Override
	public void onEnable() {
		CommandAPI.onEnable(); // Removed parameter

		// Register commands
		...
	}
}
```

### Planned TODO before merge:
- [x] Add Javadocs
- [x] Update relevant documentation
- Adress code review


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkitConfig.java
```diff
@@ -0,0 +1,42 @@
...
+	/**
+	 * Sets the CommandAPI to hook into Paper's {@link ServerResourcesReloadedEvent} when available
+	 * if true. This helps CommandAPI commands to work in datapacks after {@code /minecraft:reload}
+	 * is run.
+	 *
+	 * @param hooked whether the CommandAPI should hook into Paper's {@link ServerResourcesReloadedEvent}
+	 * @return this CommandAPIBukkitConfig
+	 */
+	public CommandAPIBukkitConfig shouldHookPaperReload(boolean hooked) {
+		this.shouldHookPaperReload = hooked;
```
JorelAli, line 34: This should error if `ServerResourcesReloadedEvent` is not present (i.e. you're running on a Spigot/CraftBukkit server). I think erroring here would be best because this is where users directly declare that they're using this hook.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin/src/main/resources/config.yml
```diff
@@ -41,6 +41,15 @@ create-dispatcher-json: false
 # implementation is actually compatible with the current Minecraft version.
 use-latest-nms-version: false
 
+# Hook into Paper's ServerResourcesReloadedEvent (default: true)
+# If "true", and the CommandAPI detects it is running on a Paper server, it will
+# hook into Paper's ServerResourcesReloadedEvent to detect when /reload is run.
```
JorelAli, line 46: This should say `/minecraft:reload` instead of `/reload` to be extra clear on what it's doing.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin/src/main/resources/config.yml
```diff
@@ -41,6 +41,15 @@ create-dispatcher-json: false
...
+# This allows the CommandAPI to automatically call its internal datapack-reloading
+# function on a server reload, allowing CommandAPI commands to be used in datapacks.
+# If you set this to false, CommandAPI commands may not work inside datapacks after
+# reloading the server.
```
JorelAli, line 50: Calling this a "server reload" is somewhat inaccurate and misleading. A server reload typically refers to `/reload` (Bukkit's plugin reloading system), whereas `/minecraft:reload` simply reloads datapacks instead. Calling it a "datapack reload" or "reloading datapacks" (you get the idea) would be much better here.

docssrc/src/config.md
```diff
@@ -120,6 +120,28 @@ use-latest-nms-version: true
 
 -----
 
+### `hook-paper-reload`
+
+Controls whether the CommandAPI hooks into the Paper-exclusive `ServerResourcesReloadedEvent` when available.
+
+When the CommandAPI detects it is running on a Paper-based server, its default behavior will be to hook into the `ServerResourcesReloadedEvent`, which triggers when `/minecraft:reload` is run. During this event, the CommandAPI runs a custom datapack reloading sequence that helps commands registered with the CommandAPI work within datapacks. See [Reloading datapacks](./internal.md#reloading-datapacks) for more information on this process.
+
+By default, this value is set to true and the CommandAPI will hook into the `ServerResourcesReloadedEvent`. If you want, you can set this to false, and the CommandAPI will not hook into this event.
```
JorelAli, line 129: > By default, this value is set to true and ...

The word "true" should be formatted as inline code.

JorelAli (changes requested):
I've pointed out a few minor changes to do with wording of some end-user-facing documentation and an assertion that I'd like added to check if users try to use the paper-specific configuration option on a non-paper server.

I've not done any manual testing, but just by static analysis of the current implementation, everything looks fairly clean and robust. I'm not going to let manual testing prevent this PR from merging, but it's something I'd like to do before 9.0.0's release.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkitConfig.java
JorelAli: The functionality before this PR was to have `hook-paper-reload` set to `true` by default. This gets checked by `checkDependencies` [here](https://github.com/willkroboth/CommandAPI/blob/dev/dev/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L145-L154) and outputs a warning to the console if it doesn't hook correctly. As such, we don't need an assertion here.

JorelAli (approved):

PR #389, opened 2022-12-29
Rename some classes

DerEchtePilz:
Sorry for the title, I didn't really find a good name...

Anyway, this PR has been inspired by [this comment on discord](https://discord.com/channels/745416925924032523/745546192661905429/1047515741978624090) and renames some interfaces (for now).

Changes in this PR:
- `AbstractExecutionInfo` has been renamed into `ExecutionInfo` (it wasn't even an abstract class, so this was kinda needed)
- `IExecutorTyped` has been renamed into `TypedExecutor`
- `IExecutorNormal` has been renamed into `NormalExecutor`
- `IExecutorResulting` has been renamed into `ResultingExecutor`

That's it for now, but I think there might be more that could be renamed. We could probably think about methods (maybe I am even willing to rename those NMS methods; those start with an `_` which doesn't follow the naming convention but I guess it's fine as long as it is consistent) or something else.


JorelAli:
I'm reluctant to name the NMS methods starting with _ - when method names are sorted alphabetically in NMS, they all appear at the top.

DerEchtePilz:
Also, should we use this PR to implement the new executors for Velocity and Sponge or should I just implement that on `JorelAli:dev/dev`?
I didn't think about them in the original PR...

JorelAli:
> Also, should we use this PR to implement the new executors for Velocity and Sponge or should I just implement that on `JorelAli:dev/dev`? I didn't think about them in the original PR...

Let's save that for implementing on `JorelAli:dev/dev`, along with the rest of our velocity/sponge development.

PR #391, opened 2023-01-01
Make the CommandAPI complain less about command registered in the plugin.yml

DerEchtePilz:
The title is not great, I know and it can be misunderstood!

### I once experienced this while using the CommandAPI:
In my plugin, I register a command permission. LuckPerms also registers a command permission.
The CommandAPI does this:
`[01:15:12 WARN]: [CommandAPI] Plugin command /permission is registered by Bukkit (LuckPerms). Did you forget to remove this from your plugin.yml file?`

### Here is what the PR does:
This PR makes it so that the CommandAPI doesn't complain about plugin commands registered in a `plugin.yml` if the `plugin.yml` doesn't belong to the plugin calling the CommandAPI.

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
@@ -387,7 +387,7 @@ public abstract class CommandAPIBukkit<Source> extends CommandAPIPlatform<Argume
 				"Plugin command /%s is registered by Bukkit (%s). Did you forget to remove this from your plugin.yml file?"
 					.formatted(commandName, pluginName));
 		} else {
-			CommandAPI.logInfo(
+			CommandAPI.logNormal(
 				"Plugin command /%s is registered by Bukkit. You may have to use /minecraft:%s to execute your command."
 					.formatted(commandName, commandName));
 		}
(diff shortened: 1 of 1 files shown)


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
```diff
@@ -378,10 +378,18 @@ public void preCommandRegistration(String commandName) {
...
+		}
+		String pluginName = pluginCommand.getPlugin().getName();
+		if (config.getPlugin().getName().equals(pluginName)) {
 			CommandAPI.logWarning(
 				"Plugin command /%s is registered by Bukkit (%s). Did you forget to remove this from your plugin.yml file?"
-					.formatted(commandName, pluginCommand.getPlugin().getName()));
+					.formatted(commandName, pluginName));
+		} else {
+			CommandAPI.logNormal(
+				"Plugin command /%s is registered by Bukkit. You may have to use /minecraft:%s to execute your command."
```
JorelAli, line 391: Is there any particular reason why we're not including the plugin name here too?

DerEchtePilz: No, there isn't really a reason. Should I add the plugin name?

JorelAli: Yes please :)

Issue #162, opened 2020-11-18
Optional arguments


JorelAli:
**Please describe your suggestion**
Optional arguments

**Please supply examples of the desired inputs and outputs**
_Normal usage:_
User enters this:
```java
new CommandAPICommand("blah")
  .withArgument(new IntegerArgument("int"))
  .withOptionalArgument(new IntegerArgument("optional"), 10)
  .executes((sender, args) -> {
    int arg1 = (int) args[0];
    int arg2 = (int) args[1]; // This is 10 if not existant
  })
  .register();
```

Basically gets converted into this:
```java
new CommandAPICommand("blah")
  .withArgument(new IntegerArgument("int"))
  .withArgument(new IntegerArgument("optional"))
  .executes((sender, args) -> {
    int arg1 = (int) args[0];
    int arg2 = (int) args[1];
  })
  .register();
```
and
```java
new CommandAPICommand("blah")
  .withArgument(new IntegerArgument("int"))
  .executes((sender, args) -> {
    int arg1 = (int) args[0];
    int arg2 = (int) args[1]; // This is 10, args.length = 2
  })
  .register();
```

_Using annotations:_
Either something like this:
```java
@Subcommand("mycommand")
@Arg(name = "int", type = IntegerArgument.class)
@OptionalArg(name = "int", type = IntegerArgument.class, optional = "10")
public void onCommand(int arg1, int arg2) {
  //...
}
```

or something like this:
```java
@Subcommand("mycommand")
@Arg(name = "int", type = IntegerArgument.class)
@Arg(name = "optional", type = IntegerArgument.class)
public void onCommand(int arg1, @Default("10") int arg2) {
  //...
}
```

Cons of this via annotations is that the value must be a string, primitive, enum, annotation or class, which is really restrictive.

SNWCreations:
Emmm, so is it really implemented?
I like this feature, but I can't find it in v7.0 :(

SNWCreations:
Emmm, so is it really implemented?
I like this feature, but I can't find it in v7.0 :(

JorelAli:
@SNWCreations No, this feature is not implemented (yet?)

SNWCreations:
> 

Emmm, that's too bad :(
If so, will it be available in v7.x? (I'm still working on 1.16.5, I need it!)

JorelAli:
@SNWCreations
[...]

PR #393, opened 2023-01-03
Implementing optional arguments

DerEchtePilz:
This PR adds optional arguments to the CommandAPI!
This has been a topic that was not only discussed serveral times on the CommandAPI discord but also exists as an issue on GitHub (#162).

This PR adds the `withOptionalArguments` method that allows users to implement, well, optional arguments.
Take this code for example:
```java
new CommandAPICommand("optionalarguments")
    .withArguments(new PlayerArgument("player"))
    .withArguments(new ItemStackArgument("item"))
    .withOptionalArguments(new IntegerArgument("amount"))
    .withOptionalArguments(new BooleanArgument("broadcast"))
    .executesPlayer(info -> {
        Player target = (Player) info.args().get("player");
        ItemStack item = (ItemStack) info.args().get("item");
        Integer amount = (Integer) info.args().get("amount");
        Boolean broadcast = (Boolean) info.args().get("broadcast");
        if (amount != null) {
            item.setAmount(amount);
        }
        target.getInventory().addItem(item);
        if (broadcast == null) {
            info.sender().sendMessage("You gave " + target.getName() + " " + item.getType().name().toLowerCase());
        } else {
            Bukkit.broadcastMessage(target.getName() + " has received " + item.getType().name());
        }
    })
    .register();
```
When executing this command in Minecraft, you have those options (for example):
```
/optionalargument DerEchtePilz minecraft:netherite_block
/optionalargument DerEchtePilz minecraft:netherite_block 5
/optionalargument DerEchtePilz minecraft:netherite_block 5 true
```

The first option puts `You gave DerEchtePilz netherite_block` in the chat and I receive one item
The second option puts `You gave DerEchtePilz netherite_block` in the chat but I receive five items
The third option puts `DerEchtePilz has received NETHERITE_BLOCK` in the chat and I receive five items

Things to add before merging:
- [x] Documentation?
- [x] Address code reviews
- [x] Add this to the Kotlin DSL
- [x] Add missing JavaDocs


commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -83,7 +84,41 @@ public Impl withArguments(List<Argument> args) {
...
+	 *
+	 * @param args A <code>List</code> that represents the arguments that this
+	 *             command can accept
+	 * @return this command builder
+	 */
+	public Impl withOptionalArguments(List<Argument> args) {
+		args.forEach(argument -> {
+			argument.setOptional(true);
+			this.arguments.add(argument);
+		});
```
JorelAli, line 104: Use a for-each loop instead of the `forEach` method:

```java
for(Argument argument : args) {
	argument.setOptional(true);
	this.arguments.add(argument);
}
```

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -83,7 +84,41 @@ public Impl withArguments(List<Argument> args) {
...
+	 *
+	 * @param args Arguments that this command can accept
+	 * @return this command builder
+	 */
+	@SafeVarargs
+	public final Impl withOptionalArguments(Argument... args) {
+		Arrays.asList(args).forEach(argument -> {
+			argument.setOptional(true);
+			this.arguments.add(argument);
+		});
```
JorelAli, line 121: Inexcusable code smell. You're converting the `Argument[]` to a list purely to access the `forEach` method when using a for-each loop does the exact same thing. You're adding a completely unnecessary amount of complexity to a really simple operation:
- One extra method call
- One constructor initialization (memory allocation to allocate the new `ArrayList` object)
- One null check to check that `args` is not null

Use a for-each loop instead of the `forEach` method:

```java
for(Argument argument : args) {
	argument.setOptional(true);
	this.arguments.add(argument);
}
```

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -212,29 +247,69 @@ public void register() {
...
 					}
 				}
 			}
 
+			// Check optional argument constraints
+			// They can only be at the end, no required argument can follow an optional argument
+			int firstOptionalArgumentIndex = -1;
+			for (int i = 0, optionalArgumentIndex = -1; i < argumentsArray.length; i++) {
+				if (argumentsArray[i].isOptional()) {
+					firstOptionalArgumentIndex = i;
```
JorelAli, line 266: This doesn't look right. You're setting `firstOptionalArgumentIndex` every time you're setting `optionalArgumentIndex`. As this iterates through the arguments array, this value will always increase - surely this is only supposed to update the first time an optional argument is found, and not update in future iterations of this loop?

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -212,29 +247,69 @@ public void register() {
...
+					optionalArgumentIndex = i;
+					continue;
```
JorelAli, line 268: This `continue` statement feels hacky. Your code flow goes:

```java
if (blah) {
  // stuff
}

if (not blah and other condition) {
  // other stuff
}
```

To catch your `not blah` condition, you've added a `continue` statement. This could be simplified using an `else` statement:

```java
if (blah) {
  // stuff
} else if (other condition) {
  // other stuff
}
```

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -212,29 +247,69 @@ public void register() {
...
+			List<Argument[]> argumentsToRegister = new ArrayList<>();
+			if (firstOptionalArgumentIndex != -1) {
+				for (int i = 0; i <= argumentsArray.length; i++) {
+					if (i == firstOptionalArgumentIndex - 1) {
+						Argument[] requiredArguments = (Argument[]) new AbstractArgument[i];
+						System.arraycopy(argumentsArray, 0, requiredArguments, 0, i);
+						argumentsToRegister.add(requiredArguments);
+						continue;
+					}
+					if (i >= firstOptionalArgumentIndex) {
```
JorelAli, line 295: This could be an `else if` statement instead of requiring a `continue` statement. (See other review comment)

JorelAli (changes requested):
Initial code review. Main review points:
- Avoid `list.forEach()` in favour of a for-each loop. Having the mindset of using `forEach()` with lambdas instead of existing features (for loops) all the time is not desirable, and has in fact degraded the quality of the code you've written. Using `forEach()` _always_ adds unnecessary performance overhead compared to a for-each loop:
  - One extra method call
  - One null check
  - One bootstrap lambda added to the Java class file (increases the file size)
  - Literally does a for-each loop under the hood

- I really like the renaming of `args` to `arguments` and `argumentsArr` to `argumentsArray`, thanks!

JorelAli (approved):

docssrc/src/arguments.md
```diff
@@ -199,26 +197,23 @@ We first register the first `/kill` command as normal:
 
 </div>
 
-Now we declare our command with arguments for our second command. Then, we can register our second command `/kill <target>` as usual:
+This gives us the ability to run both `/kill` and `/kill <target>` with the same command name "kill", but have different results based on the arguments used.
 
-<div class="multi-pre">
+In this example, we use the simpler, inline `.withOptionalArguments(Argument... arguments)` method to register our argument. There is no difference to using this method as opposed to explicitly declaring a list and using `.withOptionalArguments(List<Argument> arguments)`, so feel free to use whichever method you want!
```
JorelAli, line 202: I think this is good to mention, but it doesn't make sense in the current document, because we've not told the user that there is a `.withOptionalArguments(List<Argument> arguments)` method yet.

Typically, when I write my documentation, I like to declare the API first (if I'm not leading with an example first), then reference the API later. For example, in [Argument suggestions](https://commandapi.jorel.dev/8.7.0/argumentsuggestions.html), we show the reader the list of methods (1), then we reference them in the text (2):
image

commandapi-documentation-code/src/main/kotlin/dev/jorel/commandapi/examples/kotlin/Examples.kt
```diff
@@ -1233,21 +1233,16 @@ CommandAPICommand("mycommand")
 fun argumentkillcmd() {
 /* ANCHOR: argumentkillcmd */
 CommandAPICommand("kill")
-    .executesPlayer(PlayerCommandExecutor { player, _ ->
-        player.setHealth(0.0)
-    })
-    .register()
+    .withOptionalArguments(PlayerArgument("target"))
+    .executesPlayer(PlayerCommandExecutor { player, args ->
+        val target: Player? = args.get("target") as Player?
```
JorelAli, line 1238: Is there any reason why we're not using `args["target"]` here?

JorelAli (changes requested):
The documentation content looks good, just a few minor changes with the main content.

I think showcasing optional arguments would be much better on its own page instead of at the bottom of `arguments.md`. Can we please create a new page (e.g. `optional_arguments.md`), and place that above the "Listed arguments" section in `SUMMARY.md`.

The new page should be titled "Optional arguments" (get rid of "/Different" from the title, I don't want that anymore). I'd like some explanation about optional arguments and a block of code at the top showing the various `withOptionalArguments()` methods.

If you're struggling to meet these requirements, just let me know and I'll help out.

DerEchtePilz:
> The documentation content looks good, just a few minor changes with the main content.
> 
> I think showcasing optional arguments would be much better on its own page instead of at the bottom of `arguments.md`. Can we please create a new page (e.g. `optional_arguments.md`), and place that above the "Listed arguments" section in `SUMMARY.md`.
> 
> The new page should be titled "Optional arguments" (get rid of "/Different" from the title, I don't want that anymore). I'd like some explanation about optional arguments and a block of code at the top showing the various `withOptionalArguments()` methods.
> 
> If you're struggling to meet these requirements, just let me know and I'll help out.

- So, we remove the section `Optional/Different Arguments` from `arguments.md`?
- We are creating a new page `optional_arguments.md`?

JorelAli:
> * So, we remove the section `Optional/Different Arguments` from `arguments.md`?

Yes 👍 

> * We are creating a new page `optional_arguments.md`?

Yes 👍

JorelAli (changes requested):

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -212,29 +247,67 @@ public void register() {
...
+						Argument[] requiredArguments = (Argument[]) new AbstractArgument[i];
+						System.arraycopy(argumentsArray, 0, requiredArguments, 0, i);
+						argumentsToRegister.add(requiredArguments);
+					} else if (i >= firstOptionalArgumentIndex) {
+						Argument[] optionalArguments = (Argument[]) new AbstractArgument[i];
+						System.arraycopy(argumentsArray, 0, optionalArguments, 0, i);
+						argumentsToRegister.add(optionalArguments);
+					}
+				}
+			}
```
JorelAli: CodeFactor complains about this causing this method to be fairly complex. I initially dismissed this, but upon further inspection I think it's right. We have two parts for optional arguments here:
- Setting `firstOptionalArgumentIndex` and throwing `OptionalArgumentException`
- Generating the `argumentsToRegister` list

We could probably put both of these in one (private) function and call that instead of having these two parts.

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -212,29 +247,67 @@ public void register() {
...
+				for (int i = 0; i <= argumentsArray.length; i++) {
+					if (i == firstOptionalArgumentIndex - 1) {
+						Argument[] requiredArguments = (Argument[]) new AbstractArgument[i];
+						System.arraycopy(argumentsArray, 0, requiredArguments, 0, i);
+						argumentsToRegister.add(requiredArguments);
+					} else if (i >= firstOptionalArgumentIndex) {
+						Argument[] optionalArguments = (Argument[]) new AbstractArgument[i];
+						System.arraycopy(argumentsArray, 0, optionalArguments, 0, i);
+						argumentsToRegister.add(optionalArguments);
+					}
```
JorelAli: The contents of these two blocks are completely identical. Should they be? If so, can we make this one block instead of two?

JorelAli (changes requested):

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
DerEchtePilz: While I think one method might be good, I decided to make this two blocks because between those two block, permissions are linked to arguments.
(I hope this was understandable...)

JorelAli: You can set the permissions first, then handle optional arguments after that.

DerEchtePilz: Yeah...
Give me a moment. I'll change that!

docssrc/src/optional_arguments.md
```diff
@@ -0,0 +1,71 @@
...
+new CommandAPICommand("optional")
+    .withOptionalArguments(new PlayerArgument("target"))
+```
+
+```java
+new CommandAPICommand("optional")
+    .withArguments(new PlayerArgument("target").setOptional(true))
+```
+
+However, calling `withOptionalArguments` is safer because it makes sure that the argument is optional because of that internal call.
```
JorelAli, line 71: I think this part would look better under a new heading in a similar style to [Listed arguments](https://commandapi.jorel.dev/8.7.0/listed.html), such as:

````md
## Setting existing arguments as optional arguments

(Some description here)

```java
Argument setOptional(boolean optional);
```

(Some more text here describing that the following are identical):

```java
new CommandAPICommand("optional")
    .withOptionalArguments(new PlayerArgument("target"))
```

```java
new CommandAPICommand("optional")
    .withArguments(new PlayerArgument("target").setOptional(true))
```
````

File: commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java [hunk 11]
```diff
@@ -248,11 +292,41 @@ public void register() {
...
-		command.args = new ArrayList<>(this.args);
+		command.arguments = new ArrayList<>(this.arguments);
 		command.subcommands = new ArrayList<>(this.subcommands);
 		command.isConverted = this.isConverted;
 		return command;
 	}
 
 	protected abstract Impl newConcreteCommandAPICommand(CommandMetaData<CommandSender> metaData);
+
+	private List<Argument[]> getArgumentsToRegister(Argument[] argumentsArray, List<Argument[]> argumentsToRegister) {
```
willkroboth, line 303: Does this method need to take in the List that the arguments should be added to, then return that list?

I don't think there's any reason this wouldn't be an empty `ArrayList`, so it can probably just be created inside this method.

If you do pass in the List, you technically also don't need to return that List, because the outer function could already have a reference to it. Eg.
```java
List<?> list = new ArrayList<>();
modifyList(list);
print(list); // This list has been modified by the function
```

JorelAli (approved):

JorelAli:
In #360, there was an idea of being able to retrieve "default" values from optional arguments:

> It is useful to be able to access arguments by node name, rather than by index - especially in the case of using optional arguments:
> 
> ```java
> (CommandSender sender, Map<String, Object> args) -> {
>     String input = (String) args.get("blah");
>     int amount = (int) args.getOrDefault("amount", 1);
> }

Since we're adding all of the optional argument stuff now, it may be worth considering adding this to this PR. I think it's a good idea to add this to simplify null checks for arguments you know are going to be optional arguments:

```java
new CommandAPICommand("sayhi")
    .withOptionalArguments(new PlayerArgument("target"))
    .executesPlayer((player, args) -> {
        Player target = (Player) args.getOrDefault("target", player);
        target.sendMessage("Hi!");
    })
    .register();
```

It may also be worth considering implementing a `Supplier<?>` overload as well:

```java
new CommandAPICommand("sayhi")
    .withOptionalArguments(new PlayerArgument("target"))
    .executesPlayer((player, args) -> {
        Player target = (Player) args.getOrDefault("target", () -> getBestFriend(player));
        target.sendMessage("Hi!");
    })
    .register();
```

DerEchtePilz:
> 

You need to help me with the Supplier part though. I am not too familiar with them.
Would I implement them like this:
```java
public Object getOrDefault(int index, Supplier<?> defaultValue)
```

JorelAli:
> > 
> 
> You need to help me with the Supplier part though. I am not too familiar with them. Would I implement them like this:
> 
> ```java
> public Object getOrDefault(int index, Supplier<?> defaultValue)
> ```

Yes, that's precisely it. A `Supplier<T>` is effectively identical to `Function<Void, T>` - it's a function that doesn't have any input parameters, but returns a value. You can call the supplier using `supplier.get()`.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -48,5 +49,57 @@ public Object get(int index) {
...
+
+	/**
+	 * Returns an argument by its index
+	 *
+	 * @param index The position of this argument
+	 * @param defaultValue The Object returned if the argument is not existent
+	 * @return An argument which is placed at the given index, or the provided default value
+	 */
+	public Object getOrDefault(int index, Object defaultValue) {
+		if (args.length <= index) {
```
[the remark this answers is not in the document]
JorelAli: If `index` is negative, this also crashes.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -48,5 +49,57 @@ public Object get(int index) {
...
+
+	/**
+	 * Returns an argument by its index
+	 *
+	 * @param index The position of this argument
+	 * @param defaultValue The Object returned if the argument is not existent
+	 * @return An argument which is placed at the given index, or the provided default value
+	 */
+	public Object getOrDefault(int index, Supplier<?> defaultValue) {
+		if (args.length <= index) {
```
JorelAli, line 88: Same applies here (don't use `<=`)

docssrc/src/optional_arguments.md
```diff
@@ -77,3 +77,45 @@ new CommandAPICommand("optional")
...
+
+## Avoiding null values and exceptions
+
+Previously, we've looked at how to handle null values and how to avoid exceptions. To make all of this easier, the CommandAPI implements multiple `getOrDefault()` methods:
+
+```java
+CommandArguments getOrDefault(int index, Object defaultValue)
+CommandArguments getOrDefault(String nodeName, Object defaultValue)
+CommandArguments getOrDefault(int index, Supplier<?> defaultValue)
+CommandArguments getOrDefault(String nodeName, Supplier<?> defaultValue)
```
JorelAli, line 89: These should have a return type of `Object`, not `CommandArguments`

docssrc/src/optional_arguments.md
```diff
@@ -77,3 +77,45 @@ new CommandAPICommand("optional")
 ```
 
 However, calling `withOptionalArguments` is safer because it makes sure that the argument is optional because of that internal call.
+
+## Avoiding null values and exceptions
+
+Previously, we've looked at how to handle null values and how to avoid exceptions. To make all of this easier, the CommandAPI implements multiple `getOrDefault()` methods:
```
JorelAli, line 83: May be worth stating `CommandArguments` here, for example:

```
... the CommandAPI implements multiple `getOrDefault()` methods for `CommandArguments`:
```

JorelAli (changes requested):

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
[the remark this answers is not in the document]
DerEchtePilz: > If `args.length == index`, this crashes.

The conditional statement includes `args.length == index`?

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
DerEchtePilz: Why am I not supposed to use `<=`?

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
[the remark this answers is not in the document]
JorelAli: Yup, I'm being a dumb, ignore me 👍

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
JorelAli: I'm being dumb, ignore me

JorelAli (approved):

JorelAli (changes requested):
In `CommandArguments`, the following method can throw an ArrayIndexOutOfBoundsException:
```java
public Object get(int index) {
    return args[index];
}
```

With the introduction of optional arguments, users are now more likely to access arguments via indexes that are out of bounds. To retain feature parity with `get(String nodeName)`, please make `get(int index)` return `null` when getting an argument by position that doesn't exist. As expected, please make sure this method has `@Nullable` on it.

PR #395, opened 2023-01-05
Allow required arguments after optional arguments (Linking Edition)

DerEchtePilz:
Due to some discussions on discord, I openend this PR to present yet another method of allowing required arguments after optional arguments.
This doesn't get rid of the `OptionalArgumentException` and doesn't change the original system introduced in #393 too much.

Here is the `rate` command from #394 using linking:
```java
new CommandAPICommand("rate")
	.withArguments(
		new StringArgument("topic").setOptional(true).combineWith(new IntegerArgument("rating", 0, 10)),
		new PlayerArgument("target").setOptional(true)
	)
	.executes(info -> {
		String topic = (String) info.args().get("topic");
		if(topic == null) {
			// Honestly not sure how to justify topic being optional here,
			// causing there to be a required argument after an optional one,
			// but I'm sure an example exists, right?
			info.sender().sendMessage("You didn't give a rating");
			return;
		}

		// We know this is not null because rating is required if topic is given
		int rating = (int) info.args().get("rating");

		// The target player is optional, so give it a default here
		CommandSender target = (CommandSender) info.args().getOrDefault("target", info.sender());

		target.sendMessage("Your " + topic + " was rated: " + rating + "/10");
	})
	.register();
```
This command works exactly the same as the `rate` command in #394 but it uses linking.
The `combineWith()` method takes in any number of arguments and will lead to the `IntegerArgument` being only required if the `StringArgument` is given.
The `PlayerArgument` behaves like you would it expect it to behave.

Things left to do:
- [x] Test this. Due to the changes made to the `if`-statement, I am not sure if this does what I want it to
- [x] Possibly write documentation?
- [x] Fix some types
- [x] Maybe rename the `linkArguments` method to `combineArguments` (coming from [here](https://github.com/JorelAli/CommandAPI/pull/394#issuecomment-1372917184))
- [x] Make `combineArguments` copy permissions and requirements to the added arguments (coming from [here](https://github.com/JorelAli/CommandAPI/pull/394#issuecomment-1372917184))
- [x] Remove some imports from CommandAPIMain


DerEchtePilz:
I just thought about this and I will rename the `linkArguments` method into `combineArguments`

commandapi-core/src/main/java/dev/jorel/commandapi/arguments/AbstractArgument.java
```diff
@@ -276,6 +285,42 @@ public Impl setOptional(boolean optional) {
...
+	public final Impl combineWith(Argument... combinedArguments) {
+		for (Argument argument : combinedArguments) {
+			argument.resetRequirements();
+			argument.withRequirement(this.requirements);
+			argument.withPermission(this.permission);
+			this.combinedArguments.add(argument);
+		}
+		return instance();
+	}
+
```
willkroboth, line 323: Should the requirements and permission of the parent Argument be set right now? If I did:
```java
new StringArgument("base").combineWith(
	new StringArgument("child")
).withPermission("permission")
```
`StringArgument("child")` wouldn't get the `"permission"` permission. Same if I added a requirement after adding a combined argument. Maybe permissions and requirements could be applied to the child arguments when unpacking the combined arguments?

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -324,9 +326,16 @@ private List<Argument[]> getArgumentsToRegister(Argument[] argumentsArray) {
...
+					for (Argument argument : argumentsWithoutCombined) {
+						arguments.add(argument);
+						if (argument.hasCombinedArguments()) {
+							arguments.addAll(argument.getCombinedArguments());
+						}
+					}
+					argumentsToRegister.add(arguments.toArray((Argument[]) new AbstractArgument[0]));
 				}
 			}
 		}
```
willkroboth, line 341: This loop *can* be combined with the first for one pass total through `argumentsArray` with something like the following:
```java
private List<Argument[]> getArgumentsToRegister(Argument[] argumentsArray) {
	List<Argument[]> argumentsToRegister = new ArrayList<>();
	List<Argument> currentCommand = new ArrayList<>();

	Iterator<Argument> argumentIterator = List.of(argumentsArray).iterator();

	// Collect all required arguments, adding them as a command once finding the first optional
	while(argumentIterator.hasNext()) {
		Argument next = argumentIterator.next();
		if(next.isOptional()) {
			argumentsToRegister.add((Argument[]) currentCommand.toArray(new AbstractArgument[0]));
			currentCommand.addAll(expandCombinedArgument(next));
			break;
		}
		currentCommand.addAll(expandCombinedArgument(next));
	}

	// Collect the optional arguments, adding each one as a valid command
	while (argumentIterator.hasNext()) {
		Argument next = argumentIterator.next();
		if(!next.isOptional()) throw new OptionalArgumentException(meta.commandName); // non-optional argument after optional
		argumentsToRegister.add((Argument[]) currentCommand.toArray(new AbstractArgument[0]));
		currentCommand.addAll(expandCombinedArgument(next));
	}
	// All the arguments expanded, also handles when there are no optional arguments
	argumentsToRegister.add((Argument[]) currentCommand.toArray(new AbstractArgument[0]));
	return argumentsToRegister;
}

private List<Argument> expandCombinedArgument(Argument argument) {
	if(!argument.hasCombinedArguments()) return List.of(argument);
	List<Argument> expanded = new ArrayList<>();
	expanded.add(argument);
	for(Argument subArgument : argument.getCombinedArguments()) {
		expanded.addAll(expandCombinedArgument(argument));
	}
	return expanded;
}
```
This is still two loops, but the second starts where the first ended instead of at the beginning again. I'm not sure if the way I wrote it makes it easier to understand what the method is doing, but I think it looks cleaner and it does only look at each element of the `argumentsArray` once. This implementation does do more work when an `OptionalArgumentException` gets thrown, but developers should avoid the error being thrown anyway.

This implementation also addresses two other issues I noticed that should be considered separately.
First, what happens when a combined argument contains arguments with combined arguments? For example, if I passed this as an argument to a command:
```java
new StringArgument("a").combineWith(
	new StringArgument("b").combineWith(
		new StringArgument("c")
	)
)
```
With the current system, `StringArgument("c")` is ignored. With the implementation at the start of this comment, the recursive `expandCombinedArgument` method unpacks this argument to `[StringArgument("a"), StringArgument("b"), StringArgument("c")]`. I don't think `StringArgument("c")` should be ignored. Either it should get unpacked properly or there should be an error for having arguments combined in multiple layers like this.

Second, what happens if there are no optional arguments? I might be missing something, but it looks like the current implementation for `getArgumentsToRegister` returns an empty list when the `argumentsArray` has no optional arguments. `firstOptionalArgumentIndex = -1`, then the first loop runs. `argumentsArray[i].isOptional()` is always false, then `optionalArgumentIndex != -1` is always false, so nothing happens. `firstOptionalArgumentIndex` still equals `-1`, so the second loop gets skipped. So, the `argumentsToRegister` list gets returned with nothing in it. The implementation at the start of this comment handles this case, but it is certainly not the only way to fix this.

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -29,6 +29,7 @@
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
```
willkroboth, line 32: It looks like this import isn't used. Remove it?

commandapi-documentation-code/src/main/java/dev/jorel/commandapi/examples/java/Examples.java
```diff
@@ -1330,6 +1330,30 @@ void resultingcommandexecutor3(){
...
 
+{
+/* ANCHOR: argumentrate */
+new CommandAPICommand("rate")
+    .withOptionalArguments(new StringArgument("topic").combineWith(new IntegerArgument("rating", 0, 10)))
+    .withOptionalArguments(new PlayerArgument("target"))
+    .executes((sender, args) -> {
+        String topic = (String) args.get("topic");
+        if(topic == null) {
+            sender.sendMessage("You didn't give a rating");
```
willkroboth, line 1341: This default case could be replaced with a help message like Skepter suggested over here: https://github.com/JorelAli/CommandAPI/pull/394#issuecomment-1372559436. Maybe something like:
```java
sender.sendMessage(
	"Usage: /rate <topic> <rating> <player>(optional)",
	"Select a topic to rate, then give a rating between 0 and 10",
	"You can optionally add a player at the end to give the rating to"
)
```
And of course, make sure the same message is used in the Kotlin and DSL examples.

willkroboth (changes requested):
Overall, looks good, just a couple of bugs to iron out.

I'm still not sure that combining arguments adds enough value to be worth the extra complexity.
Pros:
- Clearer that the child argument is required when following the parent argument
- Can attach the same permission and requirement to combined arguments *(Actually, I'm not sure if this is useful. If the user doesn't have permission for the parent argument, they can't type it in. If they can't type the parent argument, the child argument is inaccessible regardless of its permission.)*

Cons:
- More complexity
    - Need to unwrap combined arguments (and handle layered combined arguments)
    - Need to enforce the `OptionalArgumentException` (relative to #394)

Commands written with #394 are almost the same as with this system. For example, if I did the following with #394:
```java
.withArguments(
	new StringArgument("A").setOptional(true),
	new StringArgument("B"),
	new StringArgument("C"),
	new StringArgument("D").setOptional(true),
	new StringArgument("E"),
	new StringArgument("F").setOptional(true)
)
```
The equivalent here would be:
```java
.withArguments(
	new StringArgument("A").setOptional(true).combineWith(
		new StringArgument("B"),
		new StringArgument("C")
	),
	new StringArgument("D").setOptional(true).combineWith(
		new StringArgument("E")
	),
	new StringArgument("F").setOptional(true)
)
```
All I see is that arguments B, C, and E are indented an extra layer. I'm not sure if that's worth the extra behind-the-scenes complexity and headaches I see in the more complicated `AbstractCommandAPICommand#getArgumentsToRegister` method. 

If the general consensus is that this PR makes developer code clearer though, I think it'll work out :).

willkroboth:
Also, I'm not sure if this is bad, but if you combine arguments the list returned by `CommandAPICommand#getArguments` is kinda wrong.
For example, these arguments:
```java
.withArguments(
	new StringArgument("A").setOptional(true).combineWith(
		new StringArgument("B"),
		new StringArgument("C")
	),
	new StringArgument("D").setOptional(true).combineWith(
		new StringArgument("E")
	),
	new StringArgument("F").setOptional(true)
)
```
Return this List:
```
[StringArgument("A"), StringArgument("D"), StringArgument("F")]
```
The length of this list of 3, even though there actually 6 arguments on this command. Without inspecting the combined arguments (kinda tricky), you wouldn't know about the arguments B, C, or E. 

Again, I don't know if this is bad since if a developer combined arguments they should know what has happened, but it's something to note.

commandapi-core/src/main/java/dev/jorel/commandapi/arguments/AbstractArgument.java
DerEchtePilz: That’s a good point. I didn‘t think about that.

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
DerEchtePilz: 1. I think this method does not make it easier to understand. I think it’s worth considering to not start a second loop inside the first one but in my opinion your solution doesn’t make it easier.
2. Nested `combineWith`methods. This was kinda intended to be like that since the `combineWith` method takes in any number of additional arguments that are all required when executing the command. 
3. It is intended for the `get ArgumentsToRegister` method to return an empty list when no optional arguments exist. I think this is acceptable since the new `combineWith` method is only intended to be used on optional arguments. If you look at the `register` method you will see that I handle this case properly so if you don’t have optional arguments, you will still have arguments in your command.

commandapi-core/src/main/java/dev/jorel/commandapi/arguments/AbstractArgument.java
DerEchtePilz: On a second thought, I think this is useful. Consider this:
- Permissions and requirements are added to the first argument
  - they will be copied over
- Permissions and requirements are added to the second argument
  - they will be overwritten
In both cases it will result in both arguments having the same permissions and requirements.

DerEchtePilz:
> Also, I'm not sure if this is bad, but if you combine arguments the list returned by `CommandAPICommand#getArguments` is kinda wrong. For example, these arguments:
> 
> ```java
> .withArguments(
> 	new StringArgument("A").setOptional(true).combineWith(
> 		new StringArgument("B"),
> 		new StringArgument("C")
> 	),
> 	new StringArgument("D").setOptional(true).combineWith(
> 		new StringArgument("E")
> 	),
> 	new StringArgument("F").setOptional(true)
> )
> ```
> 
> Return this List:
> 
> ```
> [StringArgument("A"), StringArgument("D"), StringArgument("F")]
> ```
> 
> The length of this list of 3, even though there actually 6 arguments on this command. Without inspecting the combined arguments (kinda tricky), you wouldn't know about the arguments B, C, or E.
> 
> Again, I don't know if this is bad since if a developer combined arguments they should know what has happened, but it's something to note.

I want to say, that it may not be a bad thing. The method's JavaDocs state that combined arguments are ignored until the arguments to register are built.

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -324,12 +325,29 @@ private List<Argument[]> getArgumentsToRegister(Argument[] argumentsArray) {
...
 				if (i >= firstOptionalArgumentIndex) {
-					Argument[] arguments = (Argument[]) new AbstractArgument[i];
-					System.arraycopy(argumentsArray, 0, arguments, 0, i);
-					argumentsToRegister.add(arguments);
+					List<Argument> arguments = new ArrayList<>();
+					Argument[] argumentsWithoutCombined = (Argument[]) new AbstractArgument[i];
+					System.arraycopy(argumentsArray, 0, argumentsWithoutCombined, 0, i);
+					for (Argument argument : argumentsWithoutCombined) {
+						arguments.addAll(unpackCombinedArguments(argument));
+					}
```
willkroboth, line 333: This method unpacks arguments early in the list many times when it only needs to unpack them once. To explain what I mean, consider these arguments:
```
.withArguments(
	new StringArgument("A"),
	new StringArgument("B").withPermission("permissionB").combineWith(
		// Using argument combining to share permissions
		// I think this is something someone might try, and it isn't stopped with an error
		// See a different comment to see why I think this might be useless though
		new StringArgument("C")
	)
).withOptionalArguments(
	new StringArgument("D").combineWith(
		new StringArgument("E")
	), 
	new StringArgument("F"), 
	new StringArgument("G")
)
```
After figuring out that this doesn't throw an `OptionalArgumentException` and dose include optional arguments, the second loop runs like this:
```
i = 2
new arguments
argumentsWithoutCombined = {"A", "B"}
    unpack "A" -> ["A"]
    unpack "B" -> ["B", "C"]
arguments = ["A", "B", "C"] -> register

i = 3
new arguments
argumentsWithoutCombined = {"A", "B", "D"}
    unpack "A" -> ["A"]
    unpack "B" -> ["B", "C"]
    unpack "D" -> ["D", "E"]
arguments = ["A", "B", "C", "D", "E"] -> register

i = 4
new arguments
argumentsWithoutCombined = {"A", "B", "D", "F"}
    unpack "A" -> ["A"]
    unpack "B" -> ["B", "C"]
    unpack "D" -> ["D", "E"]
    unpack "F" -> ["F"]
arguments ["A", "B", "C", "D", "E", "F"] -> register

i = 5
new arguments
argumentsWithoutCombined = {"A", "B", "D", "F", "G"}
    unpack "A" -> ["A"]
    unpack "B" -> ["B", "C"]
    unpack "D" -> ["D", "E"]
    unpack "F" -> ["F"]
    unpack "G" -> ["G"]
arguments ["A", "B", "C", "D", "E", "F", "G"] -> register
```
As you can tell, there is a lot of repeated calculations here that overall give this part of the method an `O(n^2)` time complexity. However, since unpacking an argument always gives the same result, this method can actually have an `O(n)` time complexity, like so:
```
new arguments

// handling required arguments
// If something prevented required arguments from having arguments 
// combined to them, these unpacks are also unnecessary. If that is 
// possible (eg. for permission sharing, which is questionably useful), 
// then unpacking here is necessary and expected.
unpack A -> ["A"]
arguments = ["A"]

unpack B -> ["B", "C"]
arguments = ["A", "B", "C"]

// end of required arguments
arguments = ["A", "B", "C"] -> register

// handling optional arguments
unpack "D" -> ["D", "E"]
arguments = ["A", "B", "C", "D", "E"] -> register

unpack "F" -> ["F"]
arguments = ["A", "B", "C", "D", "E", "F"] -> register

unpack "G" -> ["G"]
arguments = ["A", "B", "C", "D", "E", "F", "G"] -> register
```
If you look at when `arguments` get registered, you can see this does the same thing but in many fewer steps. Implementing this `O(n)` algorithm is an exercise left to the viewer :). The implementation I suggested [here](https://github.com/JorelAli/CommandAPI/pull/395#discussion_r1063896585) works like this, so that's what I would suggest, but other solutions are possible. The key is to remember how the previous arguments were unpacked by not resetting the `arguments` list each loop iteration.

Overall, this is similar to what happened with #301 (If you want extra reading).

commandapi-core/src/main/java/dev/jorel/commandapi/arguments/AbstractArgument.java
```diff
@@ -224,6 +225,13 @@ public final Impl withRequirement(Predicate<CommandSender> requirement) {
 		return instance();
 	}
 
+	/**
+	 * Resets the requirements for this command
+	 */
+	public final void resetRequirements() {
+		this.requirements = s -> true;
+	}
+
```
willkroboth, line 234: Should this method return an instance of the Argument? It kinda feels like a method someone might use while building an argument, though on the other hand, they should probably just give it the requirements they want and not bother resetting them. 

You did bother making this a separate public method when it could have been directly in `copyPermissionsAndRequirements` or private. If this is a method external developers should use, it'd be nice to put in in the docs.

I don't know what's right for this method though, depends on your vision for it.

willkroboth (changes requested):
Mostly some ways to optimize the `getArgumentsToRegister` method.

Note on permission (and requirement) sharing: I don't think it's actually helpful (referenced by another comment)

For example, doing this:
```java
.withArguments(
	new StringArgument("base").withPermission("permission").combineWith(
		new StringArgument("child")
	)
)
```
Arguments `base` and `child` both have the `permission` permission. However, since `child` always comes after `base`, this doesn't do anything in practice. If base had the permission, and child didn't the outcome would be the same.

If the user doesn't have `permission`, they won't be allowed to type the `base` argument, and `child` is inaccessible anyways. 
If the user does have `permission`, they can type the `base` argument. We know the user has the `permission` permission at this point, so it doesn't matter if child requires that permission as well or just has no permission.
The same logic applies to requirements and requirement sharing.

This is to say, permission sharing isn't an advantage of combining arguments since it basically does nothing.

DerEchtePilz:
> Mostly some ways to optimize the `getArgumentsToRegister` method.
> 
> Note on permission (and requirement) sharing: I don't think it's actually helpful (referenced by another comment)
> 
> For example, doing this:
> 
> ```java
> .withArguments(
> 	new StringArgument("base").withPermission("permission").combineWith(
> 		new StringArgument("child")
> 	)
> )
> ```
> 
> Arguments `base` and `child` both have the `permission` permission. However, since `child` always comes after `base`, this doesn't do anything in practice. If base had the permission, and child didn't the outcome would be the same.
> 
> If the user doesn't have `permission`, they won't be allowed to type the `base` argument, and `child` is inaccessible anyways. If the user does have `permission`, they can type the `base` argument. We know the user has the `permission` permission at this point, so it doesn't matter if child requires that permission as well or just has no permission. The same logic applies to requirements and requirement sharing.
> 
> This is to say, permission sharing isn't an advantage of combining arguments since it basically does nothing.

You miss the point that it also overwrites permissions and requirements on "child".
If `child` has a permission attached to it and `base` doesn't, you end up with a command that always is invalid.
This was also mentioned on your PR #394 
So yes, that permission sharing does indeed something useful

willkroboth:
```java
.withArguments(
	new StringArgument("base").withPermission("permission").combineWith(
		new StringArgument("child").withPermission("otherPermission")
	)
)
```
Yeah, I guess in the above case, `base` and `child` will both end up with the `"permission"` permission. However, while allowing `child` to keep `"otherPermission"` would mean that users would need both `"permission"` and `"otherPermission"` to run this command, I think that should be up to developers to decide.

I think the above example indicates that the developer wants `base` to have `"permission"` and `child` to have `"otherPermission"`. The reference to `"otherPermission"` just kinda gets deleted here without warning, which might be confusing. If they want to 'break' their command like that, they probably have a good reason? If they didn't want that, they should just not apply a permission to `child`.

DerEchtePilz:
Now, I've decided to implement your suggestion for the `getArgumentsToRegister` method since I really couldn't come up with something better.

willkroboth:
Okay, if you went with that, something to note is that I wrote it before I knew the case when there are no optional arguments was handled outside in `AbstractCommandAPICommand#register`. So basically now, the `argumentsToRegister` list will always contain some arguments, and the check for `argumentsToRegister.isEmpty()` on line 278 will always be false.

DerEchtePilz:
> Okay, if you went with that, something to note is that I wrote it before I knew the case when there are no optional arguments was handled outside in `AbstractCommandAPICommand#register`. So basically now, the `argumentsToRegister` list will always contain some arguments, and the check for `argumentsToRegister.isEmpty()` on line 278 will always be false.

Yeah, I guess this way also makes a little bit more sense since the method is called `getArgumentsToRegister`

willkroboth (approved):
The code looks good to me 👍. The concept I'm still not sure about, but I think it'll work out if Skepter approves it too.

JorelAli (changes requested):
This wasn't a requirement when this PR was originally raised, but it'll be good to follow suit with new standards. I'd like some units tests for this PR please.

JorelAli:
Bypassing the requirement for tests, we'll implement them later, but make sure they get implemented before 9.0.0's release. If any future issues arise during testing, we'll find them and we'll fix them.

PR #410, opened 2023-01-26
Adds tests for the Kotlin DSL

DerEchtePilz:


JorelAli:
Alright, calm it SonarCloud - it's fine, we don't care about code duplication. Because of the nature of this project, we're expecting code duplication all over the place.

PR #427, opened 2023-03-26
Convert Kotlin DSL executors to inline functions

Sytm:
By creating the Java Interface implementation in an inline function, the extra classes can be removed and intermediary Kotlin functions do not have to be generated.

If this style should get approved I'll do these changes to the velocity Kotlin DSL aswell.

I also noticed that there is quite a large amount of code duplication between these two DSL modules, but I can't say for certain what the best way to handle that problem in this project structure is.


DerEchtePilz:
Alright, thanks for this PR!
I didn't realize you could do this in Kotlin.

So, for this PR to get merged, this needs to be added to these files:
[CommandAPICommand DSL (Bukkit)](https://github.com/JorelAli/CommandAPI/blob/dev/dev/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-kotlin/src/main/kotlin/dev/jorel/commandapi/kotlindsl/CommandAPICommandDSL.kt)
[CommandAPICommand DSL (Velocity)](https://github.com/JorelAli/CommandAPI/blob/dev/dev/commandapi-platforms/commandapi-velocity/commandapi-velocity-kotlin/src/main/kotlin/dev/jorel/commandapi/kotlindsl/CommandAPICommandDSL.kt)
[CommandTreeDSL (Velocity)](https://github.com/JorelAli/CommandAPI/blob/dev/dev/commandapi-platforms/commandapi-velocity/commandapi-velocity-kotlin/src/main/kotlin/dev/jorel/commandapi/kotlindsl/CommandTreeDSL.kt)

Sytm:
I realized that the CommandAPICommand, CommandTree and Argument all share the common interface, which allowed me to reduce the amount of code even more

PR #430, opened 2023-04-07
Add examples for Gradle projects and the CommandAPI Kotlin-DSL as a Gradle project

Sytm:
Noteworthy is that I changed some functions in the KotlinDSL to inline even though they do not benefit from that because that way, the Kotlin DSL does not need to be shaded into the final plugin because everything is inlined.

### DIFF
--- a/docssrc/src/kotlinintro.md
+++ b/docssrc/src/kotlinintro.md
@@ -36,38 +36,6 @@ To install the DSL, you need to add the `commandapi-kotlin` dependency into your
 
 </div>
 
-Next, to shade it into your project easily, you need to add the `maven-shade-plugin`:
-
-```xml
-<build>
-    <plugins>
-        <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-shade-plugin</artifactId>
-            <version>3.3.0</version>
-            <executions>
-                <execution>
-                    <id>shade</id>
-                    <phase>package</phase>
-                    <goals>
-                        <goal>shade</goal>
-                    </goals>
-                </execution>
-            </executions>
-            <configuration>
-                <relocations>
-                    <relocation>
-                        <pattern>dev.jorel.commandapi.kotlindsl</pattern>
-                        <!-- TODO: Change this to my own package name -->
-                        <shadedPattern>my.custom.package.commandapi.kotlindsl</shadedPattern>
-                    </relocation>
-                </relocations>
-            </configuration>
-        </plugin>
-    </plugins>
-</build>
-```
-
 Next, you need to add Kotlin to your project. For this, you first need to add the dependency:
 
 ```xml
@@ -75,7 +43,7 @@ Next, you need to add Kotlin to your project. For this, you first need to add th
     <dependency>
         <groupId>org.jetbrains.kotlin</groupId>
         <artifactId>kotlin-stdlib</artifactId>
-        <version>1.8.0</version>
+        <version>1.8.20</version>
     </dependency>
 </dependencies>
 ```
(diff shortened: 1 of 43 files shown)


docssrc/src/kotlinintro.md
```diff
@@ -249,43 +167,19 @@ dependencies {
...
 
 <div class="multi-pre">
 
 ```groovy,build.gradle
-shadowJar {
-    dependencies {
-        include dependency("org.jetbrains.kotlin:kotlin-stdlib")
-    }
+kotlin {
+	jvmToolchain(16)
```
DerEchtePilz, line 176: A tab is used here. Please only use spaces in the documentation.

docssrc/src/kotlinintro.md
```diff
@@ -249,43 +167,19 @@ dependencies {
...
 }
 ```
 
 ```kotlin,build.gradle.kts
-shadowJar {
-    dependencies {
-        include dependency("org.jetbrains.kotlin:kotlin-stdlib")
-    }
+kotlin {
+	jvmToolchain(16)
```
DerEchtePilz, line 182: A tab is used here. Please only use spaces in the documentation.

DerEchtePilz (changes requested):
In general this looks good.
I've mentioned two points that need to be fixed - otherwise the tab will get displayed as 8 spaces in the documentation and that looks awful.

Also, if you're changing the Kotlin version in the documentation, please also change it in the root `pom.xml`. There is a `properties` section with `kotlin.version`. This value is used in all the Kotlin modules.

Sytm:
On the topic of tabs and spaces I edited the .editorconfig and added a spaces rule for the docssrc folder.

DerEchtePilz (approved):

Sytm:
Also the task "Build CommandAPI example Bukkit plugins in examples/ folder" looks broken (and needs changes because of the additional gradle examples)

DerEchtePilz:
I approved those changes for now - I am still not sure about the Spigot thingy. Using Paper doesn't do anything bad so I guess we'll ask @JorelAli to approve that.

Good job on the PR though!

Sytm:
I am working on this shell script atm too to also add the gradle examples

Sytm:
I could change the used toolchain in the examples to match the one used for the build to fix this failure

examples/bukkit/kotlindsl/src/main/kotlin/io/github/jorelali/Main.kt
```diff
@@ -1,15 +1,10 @@
...
 
-import dev.jorel.commandapi.CommandAPI
-import dev.jorel.commandapi.CommandAPIBukkitConfig
 import org.bukkit.plugin.java.JavaPlugin
 
 class Main : JavaPlugin() {
 
 	override fun onEnable() {
-		CommandAPI.onLoad(CommandAPIBukkitConfig(this))
-		CommandAPI.onEnable()
```
DerEchtePilz, line 11: Why did you remove these two lines? I feel like they shoud stay because the examples are supposed to present a possible way of creating a plugin using the CommandAPI.

Sytm: I removed them because they are only required when actually shading the plugin, but we are not. (https://commandapi.jorel.dev/8.8.0/shading.html).

DerEchtePilz: Thanks for pointing that out, I forgot what `commandapi-bukkit-core` is doing...

PR #436, opened 2023-04-23
Add optional parameters for all CommandTreeDSLs

Sytm:


docssrc/src/kotlindsl.md
```diff
@@ -140,7 +140,7 @@ We want to create a `/give` command with the following syntax:
 /optionalArgument give <item> <amount>
 ```
 
-As with the Java method of registering commands, with a CommandTree you can declare the command tree to include the additional argument `<amount>`, and with the CommandAPICommand you need to set the `optional` value to `true`:
+For both DSL variants you need to set the `optional` value to `true`:
```
DerEchtePilz, line 143: I would phrase this differently. Something like this would be better in my opinion:

To declare an argument as optional you need to set the `optional` value to `true`

My reason for that is that the Kotlin DSL in general is just one thing and since you changed the CommandTree methods to now also contain the `optional` value I think it might be better to not separate this here.

DerEchtePilz (changes requested):
Just a small thing I noticed, everything else looks really good!

DerEchtePilz (approved):

PR #437, opened 2023-04-23
Add support for adventure components in custom argument exceptions

booky10:
Allow using translatables, click events, hover events and other component stuff.

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/CustomArgument.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/CustomArgument.java
@@ -21,12 +21,15 @@
 package dev.jorel.commandapi.arguments;
 
 import com.mojang.brigadier.LiteralMessage;
+import com.mojang.brigadier.Message;
 import com.mojang.brigadier.context.CommandContext;
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import com.mojang.brigadier.exceptions.SimpleCommandExceptionType;
+import dev.jorel.commandapi.BukkitTooltip;
 import dev.jorel.commandapi.CommandAPIBukkit;
 import dev.jorel.commandapi.CommandAPIHandler;
 import dev.jorel.commandapi.executors.CommandArguments;
+import net.kyori.adventure.text.Component;
 import org.bukkit.command.CommandSender;
 
 import java.io.Serializable;
@@ -214,6 +217,7 @@ public class CustomArgument<T, B> extends Argument<T> {
 	@SuppressWarnings("serial")
 	public static class CustomArgumentException extends Exception {
 
+		private final Component errorComponent;
 		private final String errorMessage;
 		private final MessageBuilder errorMessageBuilder;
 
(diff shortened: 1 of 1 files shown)


JorelAli:
Looks great, thanks!

Sytm:
Wouldn't this change break on normal Spigot servers or is CommandAPI now only targeted at Paper and its forks?

`BukkitTooltip.messageFromAdventureComponent()` does not break on Spigot because if it's never called the classes aren't going to be loaded, but now that an adventure component is a class member, does adventure have to be on the classpath or only if the component constructor is called?

willkroboth:
> Wouldn't this change break on normal Spigot servers or is CommandAPI now only targeted at Paper and its forks?

Tested it, and yep, it breaks. The problem is actually with compiling though, and not with running. 

I made this simple test command in my plugin class:
```java
new CommandAPICommand("test")
        .withArguments(
                new CustomArgument<>(new IntegerArgument("even"), (info) -> {
                    int i = info.currentInput();
                    if(i%2 != 0) throw new CustomArgument.CustomArgumentException(
                            new CustomArgument.MessageBuilder("Int must be even: ")
                                    .appendFullInput().appendHere()
                    );
                    return i;
                })
        ).executes(info -> {
            info.sender().sendMessage(info.args().get(0).toString());
        })
        .register();
```
This doesn't use the `Component` constructor. If I try to compile this without a dependency for Paper, I get this error:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project CommandAPITest: Compilation failure
[ERROR] Main.java:[31,40] cannot access net.kyori.adventure.text.Component
[ERROR]   class file for net.kyori.adventure.text.Component not found
```
`[31,40]` is pointing to just before the `new` keyword in `new CustomArgument.CustomArgumentException...`. So I would guess the compiler is trying to load all of the possible constructors that could be referred to here, then gets mad about a constructor it can't understand.

If I add a dependency for Paper, the code can compile (probably because it knows about `Component`). Interestingly, I can run the resulting plugin on a Spigot server just fine:
```
>test 3
[15:31:13] [Server thread/INFO]: Int must be even: test 3<--[HERE]
>test 2
[15:31:17] [Server thread/INFO]: 2
```
So it seems Java only attempts to load the class while compiling, not running. Still, not great. I would say this change breaks the `CustomArgumentException` on Spigot servers. Probably fixable though.

Sytm:
Because the constructor takes either a dynamic Builder String or a static String/AdventureComponent/BungeeComponent, what if the CustomArgumentException now only takes the builder string or a static Brigadier Message instance that has to be provided by the developer instead (Using BukkitTooltip for example)?

booky10:
> what if the CustomArgumentException now only takes the builder string or a static Brigadier Message instance that has to be provided by the developer instead
This won't work, as brigadier is also not present at compile-time

There is no really good solution to this, I thought CommandAPI bundled adventure
It would be possible to change the Component type to Object, and then only check for Components in the "toCommandSyntax" method, but I don't think that would be a good solution

JorelAli:
While not the most elegant solution, I using static methods may solve this:

```java
static CustomArgumentException of(Component errorMessage)
```

booky10:
This still doesn't work, as the `of()` method is also declared multiple times like the constructor was previously.

The compiler can't find the Component class while trying to figure out which method should be used, so the `of()` methods should be renamed (like it was done previously with the fail methods).

JorelAli:
Really? Worked when I tried this...

booky10:
Class:

```java
import dev.jorel.commandapi.arguments.CustomArgument;
import net.md_5.bungee.api.chat.TextComponent;

public class Testing {

    public static void main(String[] args) {
        CustomArgument.CustomArgumentException.of("");
        CustomArgument.CustomArgumentException.of(TextComponent.fromLegacyText(""));
        CustomArgument.CustomArgumentException.of(new CustomArgument.MessageBuilder(""));
    }
}
```

Compiler output:

```
[project path]/src/main/java/Testing.java:7: error: cannot access Component
        CustomArgument.CustomArgumentException.of("");
                                              ^
  class file for net.kyori.adventure.text.Component not found
```

Using `org.spigotmc:spigot-api:1.19.4-R0.1-SNAPSHOT` and `dev.jorel:commandapi-bukkit-core:9.0.1` (from maven local) as dependencies.

JorelAli:
Amended on `dev/dev` to have different names for the static factory methods.

PR #459, opened 2023-06-08
Add the option to retrieve raw arguments

DerEchtePilz:
I think this needs a little bit more work but I am still opening the PR now

ToDo's:
- [x] Add tests
- [x] Write documentation
  - [x] new page
  - [x] link every appearance of `CommandArguments` to that new page
  - [x] Update `upgrading.md`
  - [x] Update `intro.md`
  - [x] Update `optional_arguments.md`
  - [x] Add examples
- [x] Remove `@Deprecated` annotation from the `CommandArguments#getOrDefaultUnchecked` methods because the `CommandArguments#getOptionalUnchecked` methods require you to provide a type or cast it which is not great
- [x] Update JavaDocs for all of the getRaw methods. They should explain what a "raw argument" is
- [x] Update every example to use the node name to access arguments instead of the index


commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -274,16 +274,28 @@ CommandArguments argsToCommandArgs(CommandContext<Source> cmdCtx, Argument[] arg
...
+		List<String> rawArguments = new ArrayList<>();
+
+		// LinkedHashMap for raw arguments
+		Map<String, String> rawArgumentsMap = new LinkedHashMap<>();
+
 		// Populate array
 		for (Argument argument : args) {
 			if (argument.isListed()) {
-				Object parsedArgument = parseArgument(cmdCtx, argument.getNodeName(), argument, new CommandArguments(argList.toArray(), argsMap, "/" + cmdCtx.getInput()));
+				Object parsedArgument = parseArgument(cmdCtx, argument.getNodeName(), argument, new CommandArguments(argList.toArray(), argsMap, rawArguments.toArray(String[]::new), rawArgumentsMap, "/" + cmdCtx.getInput()));
```
JorelAli, line 286: Don't use `toArray(String[]::new)`, use `toArray(new String[0])` instead.

JorelAli (changes requested):
Don't use `toArray(String[]::new)`, use `toArray(new String[0])` instead when converting from a `List<String>` (especially an `ArrayList<String>`) to a `String[]`

JorelAli (approved):

JorelAli:
I'm not entirely sold on the naming convention of "raw", but I don't have any better ideas.

DerEchtePilz:
Well, I thought that if something wasn't dealt with already, it exists in its raw form...
I can totally think about the name again but in my opinion it kinda fits.

JorelAli:
> Well, I thought that if something wasn't dealt with already, it exists in its raw form... I can totally think about the name again but in my opinion it kinda fits.

If you're happy with it, I'm happy with it. I got nothing. Just have to make sure it's clear in the documentation what we mean by a "raw" argument as not to cause any misunderstandings.

JorelAli:
Other than writing documentation for this, is this feature-complete?

DerEchtePilz:
Yeah, I think so.
I would also add some basic tests somewhere but I put tests on the same level as documentation writing so that will happen in one step.
I would wait to merge this (if that's what you want) because then I'll probably forget about documentation xD

JorelAli:
I'm only willing to merge this when documentation has been written. Good shout for adding tests, I forgot about that!

DerEchtePilz:
About the documentation, currently I am not very motivated to do this (partly because I want to add a complete new page for the `CommandArguments` class) so I will get to it at some point later.
This isn't urgent, right (so I can take the time I need)?

JorelAli:
> currently I am not very motivated to do this
Fair enough, I've still not even started working on the overhauled annotation system yet. These things just happen!

> This isn't urgent, right (so I can take the time I need)?
Yes, this is not an urgent feature, feel free to take all the time you need.

willkroboth (commented):
Just some quick looks at the code. Only little tweaks that don't really have a big impact either way, so you could totally just resolve the comments and not use them if you think the current way is fine.

Sorry, I couldn't get to the documentation because I have to go now. I can just suggest looking through the GitHub Actions failures, which is complaining about the spacing between some of the lines (all lines should have 1 empty line between them).

docssrc/src/optional_arguments.md
```diff
@@ -78,24 +78,19 @@ However, calling `withOptionalArguments` is safer because it makes sure that the
...
 ```java
+Object getOrDefault(int index, Object defaultValue);
+Object getOrDefault(int index, Supplier<?> defaultValue);
+Object getOrDefault(String nodeName, Object defaultValue);
+Object getOrDefault(String nodeName, Supplier<?> defaultValue);
 Optional<Object> getOptional(int index)
 Optional<Object> getOptional(String nodeName)
 ```
 
+The examples will be using the `getOptional` methods but there is no downside of using the `getOrDefault` methods.
```
JorelAli, line 92: Nice to mention this, but I think it's worth stating what the `getOrDefault` methods _do_ instead of brushing them aside with no explanation.

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+
+The `CommandArguments` class was introduced in CommandAPI 9.0.0 and provides a much more powerful way of accessing arguments than just an array of arguments which existed until 9.0.0.
+
+While the argument array just gives the possibility to access the arguments via the array notation (`args[0]`), the `CommandArguments` class offers much more, including:
+
+- [Access the inner structure directly](#access-the-inner-structure-directly)
+- [What terms are used?](#what-terms-are-used)
+  - [`nodeName`](#nodename)
+  - [`raw argument`](#raw-argument)
+  - [`unsafe argument`](#unsafe-argument)
```
JorelAli, line 11: This doesn't make sense in the context of the sentence above it. The sentence above states "... class offers much more, including:" and then this bullet point says "What terms are used?".

It looks like you're trying to make a table of contents. This isn't necessary, assuming the table of contents renders properly on the side of the page (which may not because the code to do that is super old and was when I didn't know how to write JavaScript):

image

It may be better to just summarize the top-level sections, e.g:

- Accessing arguments by index (previous behavior)
- Accessing arguments by their node name
- Accessing raw arguments
- Accessing arguments with inferred type-checking

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+- a `String` which holds the full input
+
+Although not recommended, it is possible to access these fields directly with methods the `CommandArguments` class provides:
+
+```java
+Object[] args();                  // Returns the argument array
+Map<String, Object> argsMap();    // Returns an unmodifiable map containing the arguments mapped to their node names
+String[] rawArgs();               // Returns the raw argument array
+Map<String, String> rawArgsMap(); // Returns an unmodifiable map containing the raw arguments mapped to their node names
+String getFullInput();               // Returns the full command string (including the /)
```
JorelAli, line 43: The spacing before the last comment is off

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+
+To provide arguments, the `CommandArguments` class stores:
+
+- a `Object[]` of parsed arguments
+- a `Map<String, Object>` of parsed arguments mapped to their node names
+- a `String[]` of raw arguments
+- a `Map<String, String>` of raw arguments mapped to their node names
+- a `String` which holds the full input
+
+Although not recommended, it is possible to access these fields directly with methods the `CommandArguments` class provides:
```
JorelAli, line 36: Please omit this sentence. This is a highly touchy subject.

(For context, accessing record fields using methods is _the recommended way_ to access record fields _under the assumption_ that you're enforcing OOP which Java is seen as a "golden standard" for OOP principles. We don't override any of these fields, so we don't care.)

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Access the inner structure directly
+
+To provide arguments, the `CommandArguments` class stores:
+
+- a `Object[]` of parsed arguments
+- a `Map<String, Object>` of parsed arguments mapped to their node names
+- a `String[]` of raw arguments
+- a `Map<String, String>` of raw arguments mapped to their node names
+- a `String` which holds the full input
```
JorelAli, line 34: I don't really like how this is laid out. I think it may be better to give a full explanation in a block rather than a few words which could lead to ambiguity or otherwise. An example of a suitable layout may be this, utilizing subheadings to collate similar methods (such as a section for parsed arguments, a section for raw arguments, a section for other utility methods): https://commandapi.jorel.dev/9.0.3/commandregistration.html#commandapicommand-methods

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+
+```java
+Object[] args();                  // Returns the argument array
+Map<String, Object> argsMap();    // Returns an unmodifiable map containing the arguments mapped to their node names
+String[] rawArgs();               // Returns the raw argument array
+Map<String, String> rawArgsMap(); // Returns an unmodifiable map containing the raw arguments mapped to their node names
+String getFullInput();               // Returns the full command string (including the /)
+```
+
+Additionally, the `CommandArguments` class has one more method that isn't directly backed by a field and returns the amount of arguments for a command:
```
JorelAli, line 46: Don't have to state this. See the above comment about restructuring this section with subheadings - this would be a good candidate for "utility methods"

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+
+```java
+int count();
+```
+
+While these methods can be used to access arguments, it may be safer to use the other methods the `CommandArguments` class provides to access arguments.
```
JorelAli, line 52: Omit this sentence.

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+
+### `nodeName`
+
+The `nodeName` is set when initializing an argument. For example:
+
+```java
+new StringArgument("string") 
+```
+
+The `nodeName` here would be `string`.
```
JorelAli, line 68: This should probably not appear in this page. I think it may be better to appear in `arguments.md` where we mention "node" for the first time.

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+When [accessing the raw arguments](#access-raw-arguments) of this command there are `15.3` and `@e` available as `String`s.
+
+However, when [accessing the arguments](#access-arguments) of this command there is `15.3` available as `double` and `@e` available as `Collection<Entity>`.
+
+### `unsafe argument`
+
+When [accessing arguments](#access-arguments) you need to cast the `Object` returned by these methods to the type the argument returns. More about casting arguments [here](./arguments.md#argument-casting).
+
+However, the `CommandArguments` class provides a way to remove the need to cast the arguments which is referred to as `unsafe arguments` in this page.
+
```
JorelAli, line 89: I'd rather introduce these terms under the relevant sections (e.g. put raw argument's explanation in the raw argument section below and put the unsafe argument explanation in the unsafe argument section below).

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+When [accessing the raw arguments](#access-raw-arguments) of this command there are `15.3` and `@e` available as `String`s.
+
+However, when [accessing the arguments](#access-arguments) of this command there is `15.3` available as `double` and `@e` available as `Collection<Entity>`.
+
+### `unsafe argument`
+
+When [accessing arguments](#access-arguments) you need to cast the `Object` returned by these methods to the type the argument returns. More about casting arguments [here](./arguments.md#argument-casting).
+
+However, the `CommandArguments` class provides a way to remove the need to cast the arguments which is referred to as `unsafe arguments` in this page.
```
JorelAli, line 88: Unsafe arguments should have an example (similar to how the raw arguments has an example) showcasing how it differs to accessing arguments normally. It should also point out the issues with it (e.g. doesn't work with Kotlin's inference system).

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -179,6 +191,111 @@ public Optional<Object> getOptional(String nodeName) {
...
+		} else {
+			return rawArgs[index];
+		}
+	}
+
+	/**
+	 * Returns a raw argument by its node name
+	 *
+	 * @param nodeName The node name of this argument. This was set when initializing an argument
+	 * @return A raw argument which has the given node name. Can be null if <code>nodeName</code> was not found.
```
JorelAli, line 214: This description is inconsistent with the `getRaw` method that uses `{@code null}` in it. It would probably be better to use `{@code null}` here as well.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -179,6 +191,111 @@ public Optional<Object> getOptional(String nodeName) {
...
+	 */
+	@Nullable
+	public String getRaw(String nodeName) {
+		return rawArgsMap.get(nodeName);
+	}
+
+	/**
+	 * Returns a raw argument by its index
```
JorelAli, line 222: A description of what a "raw argument" is needed here, otherwise anyone reading this JavaDoc will have no idea what this means. (Same applies to other raw-argument-accessing methods)

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -179,6 +191,111 @@ public Optional<Object> getOptional(String nodeName) {
...
+			return rawArgs[index];
+		}
+	}
+
+	/**
+	 * Returns a raw argument by its node name
+	 *
+	 * @param nodeName     The node name of this argument. This was set when initializing an argument
+	 * @param defaultValue The String returned if the raw argument was not found.
+	 * @return A raw argument with the specified node name or the provided default value
```
JorelAli, line 241: "A raw argument with the specified node name or the provided default value" sounds misleading. This could be misinterpreted as:

- Returns `nodeName` or `defaultValue`
- Returns the raw argument value (this is what it actually does)

This may be improved by including a definition of a "raw argument" in this JavaDoc.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -179,6 +191,111 @@ public Optional<Object> getOptional(String nodeName) {
...
+	 * @param nodeName     The node name of this raw argument. This was set when initializing an argument
+	 * @param defaultValue The String returned if the raw argument was not found.
+	 * @return A raw argument with the specified node name or the provided default value
+	 */
+	public String getOrDefaultRaw(String nodeName, Supplier<String> defaultValue) {
+		return rawArgsMap.getOrDefault(nodeName, defaultValue.get());
+	}
+
+	/**
+	 * Returns an <code>Optional</code> holding the raw argument by its index
```
JorelAli, line 274: `Optional` shouldn't be wrapped in `<code>`, it should be in a `@link` instead (e.g. `{@link Optional}`)

JorelAli (changes requested):
Documentation reviewed as requested. I've requested a couple changes, notably:
- A bit of re-organizing of the `CommandArguments` page by removing the table-of-contents and the "What terms are used?" sections
- Changing the format of "Access the inner structure directly" to match the way other pages in the documentation explain similar concepts

There's a lot of repetition with the node name and accessing by index under each section! After a long think about it, I'm going to approve that - I don't think it needs changing, but ensure that when it comes to writing examples you keep them concise - an example isn't necessary for every single method and once you've given an example of one concept you don't have to repeat it for every one (for example, you can write an example for `get(String)` and `get(int)`, an example for `getOrDefault(String, Object/Supplier)` and `getOptional(String)` in the "Access arguments" section. In the "access raw arguments" section, an example of just accessing `getRaw(String)` would be sufficient and likewise with the "access unsafe arguments" section - an example of `getUnchecked(String)` would be sufficient).

An explanation at the top of "access raw arguments" is necessary to let users know what we're talking about (refactoring the explanation in "What terms are used" to this section would make a lot of sense here). Likewise with unsafe arguments.

Hopefully this review isn't too long!

docssrc/src/commandarguments.md
```diff
@@ -0,0 +1,202 @@
...
+  - [Access arguments by node name](#access-arguments-by-node-name)
+  - [Access arguments by index](#access-arguments-by-index)
+- [Access raw arguments](#access-raw-arguments)
+  - [Access raw arguments by node name](#access-raw-arguments-by-node-name)
+  - [Access raw arguments by index](#access-raw-arguments-by-index)
+- [Access unsafe arguments](#access-unsafe-arguments)
+  - [Access arguments by node name](#access-arguments-by-node-name-1)
+  - [Access arguments by index](#access-arguments-by-index-1)
+
+ATTENTION!!!!! Every executable and generated documentation has to be deleted before committing anything!
```
JorelAli, line 22: And of course, remember to remove this before we're ready to merge this PR!

docssrc/src/optional_arguments.md
DerEchtePilz: Wouldn't it make more sense to only mention the `getOrDefault` here and explain it in the `CommandArguments` page?
That page is the first one in the `Command arguments` section anyway, so only mentioning it here is probably fine?

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
DerEchtePilz: I am going to resolve this because I added the definition of a "raw argument" but if there is something else you want me to add/change, I'll do it.

docssrc/src/optional_arguments.md
JorelAli: > That page is the first one in the `Command arguments` section anyway

Good point.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -10,49 +10,18 @@
...
-	 * @param fullInput The raw command a player has entered
-	 */
-	public CommandArguments(Object[] args, Map<String, Object> argsMap, String[] rawArgs, Map<String, String> rawArgsMap, String fullInput) {
-		this.args = args;
-		this.rawArgs = rawArgs;
-		this.argsMap = argsMap;
-		this.rawArgsMap = rawArgsMap;
-		this.fullInput = fullInput;
-	}
+public record CommandArguments(Object[] args, Map<String, Object> argsMap, String[] rawArgs, Map<String, String> rawArgsMap, String fullInput) {
```
JorelAli, line 21: JavaDocs for the parameters should be added to the individual arguments. See [SuggestionInfo](https://github.com/JorelAli/CommandAPI/blob/55fd6ffa45ca6f692a4237cbfada0247fc8a82c2/commandapi-core/src/main/java/dev/jorel/commandapi/SuggestionInfo.java#L37) as an example, where each parameter has JavaDocs attached to them.

I've discussed this in the past, but certain IDEs (Eclipse) doesn't generate JavaDocs for the function implementations of record fields (e.g. `args()` or `argsMap()`) unless JavaDocs are added to the given parameters.

JorelAli (changes requested):
Other than the issue with JavaDocs for the `CommandArguments` record, we're ready to merge.

PR #461, opened 2023-06-10
Work on MultiLiteralArgument constructors

DerEchtePilz:
This PR aims to add a constructor for the `MultiLiteralArgument` that was supposed to be released in version 9.0.2.
Somehow I didn't think about that and I also had weird issues when implementing that so I went with a constructor that takes in the node name and a list of literals.
The constructor that was added in 9.0.2 looked like this:
```java
public MultiLiteralArgument(String nodeName, List<String> literals)
```

Another problem in 9.0.2 is the removal of the old constructor `MultiLiteralArgument(String... literals)`. While this was supposed to be deprecated, it ended up being replaced with this constructor:
```java
public MultiLiteralArgument(String[] literals)
```
This of course breaks any usages of the MultiLiteralArgument.

This PR deprecates the above mentioned constructors and add this new constructor that, as I mentioned at the top, was supposed to be existing already:
```java
public MultiLiteralArgument(String nodeName, String... literals)
```

**Note to myself: This is now for 9.0.4 because 9.0.3 was recently released!**

ToDo's before merging:
- [x] Replace any deprecated `MultiLiteralArgument` constructors with the new one. This includes:
  - [x] Constructors in tests
  - [x] Constructors in documentation examples
  - [x] Constructors in code examples
  - [x] Constructors in code
- [x] Update documentation to mention the new constructor
- [x] Double-check if `@Deprecated` annotation is placed on the correct methods (Java API, Kotlin DSL)
- [x] Update versions in `@Deprecated` annotations (Java API, Kotlin DSL)
- [ ] Any other updates that may be necessary but I have forgotten now


examples/bukkit/commandtrees/src/main/java/io/github/jorelali/BetterBossBarsCommand.java
```diff
@@ -66,7 +66,7 @@ public void registerBetterBossBarCommand() {
 						)
 					)
 					.then(new LiteralArgument("style")
-						.then(new MultiLiteralArgument("styles", List.of("notched_6", "notched_10", "notched_12", "notched_20", "progress"))
+						.then(new MultiLiteralArgument("styles", "notched_6", "notched_10", "notched_12", "notched_20", "progress")
```
JorelAli, line 69: After seeing this change, I'm no longer happy with this example plugin anymore. Given the changes that we've made to support named nodes, we're still accessing arguments by index in our executors:

```java
NamespacedKey id = (NamespacedKey) args.get(0);
String style = (String) args.get(1);
```

Would it be unreasonable to try and push for accessing arguments by their node names as the default convention?

JorelAli (approved):

PR #468, opened 2023-06-28
Adds resolver to BukkitCommandSender for unknown senders.

469512345:
This PR simply adds a utility method for wrapping an unknown bukkit CommandSender object into a BukkitCommandSender.

```java
static @Nullable BukkitCommandSender<?> fromUnknownCommandSender(CommandSender source) {
		if(source instanceof BlockCommandSender blockCommandSender) {
			return new BukkitBlockCommandSender(blockCommandSender);
		}
		if(source instanceof ConsoleCommandSender consoleCommandSender) {
			return new BukkitConsoleCommandSender(consoleCommandSender);
		}
		//Player needs to go before entity as player extends entity
		if(source instanceof Player player) {
			return new BukkitPlayer(player);
		}
		if(source instanceof Entity entity) {
			return new BukkitEntity(entity);
		}
		if(source instanceof NativeProxyCommandSender nativeProxyCommandSender) {
			return new BukkitNativeProxyCommandSender(nativeProxyCommandSender);
		}
		if(source instanceof ProxiedCommandSender proxiedCommandSender) {
			return new BukkitProxiedCommandSender(proxiedCommandSender);
		}
		return null;
	}
```

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/commandsenders/BukkitCommandSender.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/commandsenders/BukkitCommandSender.java
@@ -1,6 +1,45 @@
 package dev.jorel.commandapi.commandsenders;
 
+import dev.jorel.commandapi.wrappers.NativeProxyCommandSender;
+import org.bukkit.command.BlockCommandSender;
 import org.bukkit.command.CommandSender;
+import org.bukkit.command.ConsoleCommandSender;
+import org.bukkit.command.ProxiedCommandSender;
+import org.bukkit.entity.Entity;
+import org.bukkit.entity.Player;
+import org.jetbrains.annotations.Nullable;
 
 public interface BukkitCommandSender<Source extends CommandSender> extends AbstractCommandSender<Source> {
+
+	/**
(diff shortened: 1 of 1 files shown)


DerEchtePilz:
So, seems a bit redundant as this already exists in the CommandAPIBukkit class:
https://github.com/JorelAli/CommandAPI/blob/master/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L386-L400

willkroboth:
> Would be good to see the original method annotated as @nullable though for null safety.

Hm, I guess if it can return null it should be marked @Nullable. However, maybe it should throw an exception instead like [VanillaCommandWrapper#getListener](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/command/VanillaCommandWrapper.java#86) does. Spigot doesn’t [want](https://hub.spigotmc.org/jira/browse/SPIGOT-7023?focusedId=41009&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-41009) there to be any CommandSender subclasses beyond their own, so the exception should never be thrown any way. 

If a CommandSender can run Vanilla (and CommandAPICommand) commands, then CommandAPIBukkit#wrapCommandSender will work as well. If not, I think it makes sense to throw an exception the same as VanillaCommandWrapper.

PR #471, opened 2023-07-21
Replace the NMS implementation of resendPackets with Player.updateCommands

Sytm:
`Player.updateCommands` has been in the Bukkit API since August 2018 and performs the same actions as the NMS implementations (with an additional check that the player is actually connected to the server).

I also tested it and to no surprise it works like before


willkroboth (approved):

DerEchtePilz (approved):

JorelAli:
Worth putting a mention in the global changelog so we can keep track of this change for the next release.

PR #482, opened 2023-08-17
Add Kotlin DSL support for delegations

DerEchtePilz:
As suggested by Sparky in the [CommandAPI Discord](https://discord.com/channels/745416925924032523/745418987948736543/1140692747221737582), this PR is adding support for delegations in the Kotlin DSL.

This would allow retrieving arguments like this:
```kotlin
CommandAPICommand("cmd")
    .withArguments(StringArgument("string"))
    .executes(CommandExecutor { _, args -> 
        val string: String by args
    })
    .register()
```
Or, in the Kotlin DSL:
```kotlin
commandAPICommand("cmd") {
    stringArgument("string")
    anyExecutor { _, args -> 
        val string: String by args
    }
}
```

This PR restructures the Kotlin DSL modules which are now no longer included in the `commandapi-bukkit` and `commandapi-velocity` modules. Instead, the `commandapi-kotlin` module has been added to the root of the project which contains three modules: `commandapi-core-kotlin`, `commandapi-bukkit-kotlin` and `commandapi-velocity-kotlin`.
This results in this structure:
- `commandapi-core`
- `commandapi-kotlin`
  - `commandapi-core-kotlin`
  - `commandapi-bukkit-kotlin`
  - `commandapi-velocity-kotlin`

To still be able to build all modules that belong to a specific platform, these modules have also been added to the `commandapi-platforms` `pom.xml`.

ToDo's:
- [x] Add an entry to the changelog
- [x] Write some documentation
  - [x] Add the documentation to `kotlindsl.md`
  - [x] Update documentation changelog in `intro.md`


JorelAli:
> - commandapi-core
> - commandapi-kotlin
>   - commandapi-core-kotlin
>   - commandapi-bukkit-kotlin
>   - commandapi-velocity-kotlin

Would it not make more sense to name them `commandapi-kotlin-core`, `commandapi-kotlin-bukkit` and `commandapi-kotlin-velocity`?

JorelAli (approved):

willkroboth (changes requested):
Just some notes about the `pom.xml` files.

I don't know enough to judge the Kotlin part, but it probably works. Maybe there could be some tests added for the feature.

DerEchtePilz:
> Just some notes about the `pom.xml` files.
> 
> I don't know enough to judge the Kotlin part, but it probably works. Maybe there could be some tests added for the feature.

@willkroboth 
This PR isn't done yet, I already planned to write tests and some documentation for it because it's also possible to use this while not having the Kotlin DSL modules `commandapi-bukkit-kotlin` and `commandapi-velocity-kotlin` as a dependency.

Thanks about the project version thingy though, I totally forgot that!

About the structure and the file paths. Given that it works I don't think the profile thing should be added to `commandapi-kotlin`. The goal of this was to have all the Kotlin stuff organized (hence the removal of `commandapi-[platform]-kotlin` from the `commandapi-[platform]` module) and I wanted no code repetition if possible which is why I added `commandapi-kotlin` in the first place.
Adding the profiles to `commandapi-kotlin` in my opinion defeats the last point as there would be essentially the same functionality at two places. The way it currently is makes it more clear what belongs to `Platform.Bukkit` and what belongs to `Platform.Velocity`, at least in my opinion.

I am also planning on doing some more testing with the modules to see how they behave when set up differently but that's nothing that I want to get too deep into now.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-kotlin-test/src/test/kotlin/dev/jorel/commandapi/test/dsltests/DelegationTests.kt
```diff
@@ -0,0 +1,79 @@
...
+				val string: String by args
+				val target: Player by args
+			}
+		}
+
+		val player: PlayerMock = server.addPlayer("Player1")
+
+		assertDoesNotThrow {
+			server.dispatchCommand(player, "test testString Player1")
+		}
```
willkroboth, line 76: Perhaps instead of just saying this call does not throw any exceptions, you could also test that the `string` and `target` variables were set to the expected values. Two `Mut#set` in the executor and two `assertEquals` should do the trick. That would just make sure that the argument values aren't being swapped or set to null, or something crazy like that. They probably aren't wrong, but it's good to check.

willkroboth (changes requested):
Just one comment on the tests.

I still think it makes sense to put

```xml
<modules>
	<module>commandapi-core-kotlin</module>
</modules>

<profiles>
	<profile>
		<id>Platform.Bukkit</id>
		<modules>
			<module>commandapi-bukkit-kotlin</module>
		</modules>
	</profile>
	<profile>
		<id>Platform.Velocity</id>
		<modules>
			<module>commandapi-velocity-kotlin</module>
		</modules>
	</profile>
</profiles>
```

into the `pom.xml` of `commandapi-kotlin`. If you want all the Kotlin stuff in the same place, that would keep all the Kotlin stuff inside the `kotlin` module instead of putting it in `commandapi-platforms`. 

If you want it to be more clear what belongs to Bukkit/Velocity, you should put `commandapi-bukkit-kotlin` inside `commandapi-bukkit`'s pom and `commandapi-velocity-kotlin` inside `commandapi-velocity`'s pom, like it was before this PR. You do have to do the path traversal, which I think makes this slightly worse, but the purpose of the `commandapi-[platform]` modules are to store the platform-specific modules.

However, I suppose building the CommandAPI is functionally the same in any of these ways, so I'm not requesting changes for that anymore.

willkroboth (approved):
Looks good to me, to the extent that I can judge Kotlin code :P. I like the extra check that the exception message is right.

Still not sure about this: https://github.com/JorelAli/CommandAPI/pull/482/files#r1297831688, but oh well.

I'm guessing docs are on the TODO list?

docssrc/src/kotlindsl.md
```diff
@@ -127,6 +127,36 @@ This works similar to how argument behaviour is modified in a CommandTree:
 
 </div>
 
+## Delegations
+
+The Kotlin DSL offers an additional way to access arguments: delegations. With delegations, there are two possible dependencies you can use:
```
JorelAli, line 132: If my understanding is correct, isn't the technical term for what we've implemented here "delegated properties"? According to the [Kotlin documentation for Delegations](https://kotlinlang.org/docs/delegation.html), a delegation is a design pattern for inheritance, whereas the [Kotlin documentation for Delegated properties](https://kotlinlang.org/docs/delegated-properties.html) looks more akin to what we've implemented.

That said, a link to the Kotlin documentation for this pattern would be appreciated. I presume this sort of thing is much more common in the Kotlin developer scene, but it would be a good idea to point new Kotlin developers to the corresponding documentation.

PR #499, opened 2023-10-19
Fix 'advenure' typos

dkim19375:
Not sure if I need to put `since = "..."`

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/BukkitTooltip.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/BukkitTooltip.java
@@ -83,31 +83,31 @@ public class BukkitTooltip<S> extends Tooltip<S> {
 	 * @param tooltipGenerator function which returns a formatted tooltip for the suggestion, an adventure {@link Component}
 	 * @param suggestions array of suggestions to provide to the user
 	 *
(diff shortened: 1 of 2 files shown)


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/BukkitTooltip.java
```diff
@@ -99,13 +117,29 @@ public static <S> Collection<Tooltip<S>> generateAdvenureComponents(Function<S,
 	 * @param tooltipGenerator function which returns a formatted tooltip for the suggestion, an adventure {@link Component}
 	 * @param suggestions collection of suggestions to provide to the user
 	 *
+     * @deprecated This method has been deprecated in favour of {@link BukkitTooltip#generateAdventureComponents(Function, Collection)}
 	 * @return a collection of {@link Tooltip <S>} objects from the provided suggestions, with the generated formatted
 	 * 	tooltips
 	 */
 	public static <S> Collection<Tooltip<S>> generateAdvenureComponents(Function<S, Component> tooltipGenerator, Collection<S> suggestions) {
```
JorelAli, line 124: This needs a `@Deprecated` annotation

JorelAli (changes requested):

JorelAli:
I think it's worth putting `since = "9.3.0"`, even if it's not a major change and is going to be removed, it's good for future developers that are maintaining this software (knowing when something was deprecated helps give an idea when to remove it), as well as developers using the CommandAPI (knowing when it was made known that this is a dodgy method name)

PR #501, opened 2023-10-21
Run command tree modify tasks with `Commands.COMMAND_SENDING_POOL` on Paper

### LINKED ISSUE #494: ConcurrentModificationException while unregistering plugin in onDisable
### CommandAPI version

9.2.0

### Minecraft version

1.19.4

### Are you shading the CommandAPI?

Yes

### What I did

Running Paper git-Paper-524 (MC: 1.19.4) (Git: 9846d0d)

Using the below test plugin which shades/shadows CommandAPI, and a plugin which allows unloading individual plugins (such as PlugmanX or ServerUtils), start the server as normal, then unload our test plugin.

```java
@Override
public void onLoad() {
    CommandAPI.onLoad(
        CommandAPIBukkitConfig(this)
            .silentLogs(false)
            .verboseOutput(true)
    );
}

@Override
public void onEnable() {
    CommandAPI.onEnable();

    new CommandAPICommand("stuff")
        // This is where the command would go if I could find a case where it reliably breaks.
        .register();
}

@Override
public void onDisable() {
    CommandAPI.unregister("stuff");
    CommandAPI.onDisable();
}
```

### What actually happened

The command was unregistered, but doing it left the following stack trace in the server logs:
```
[22:54:00 INFO]: [CommandAPI] Unregistering command /stuff
[22:54:00 ERROR]: Caught previously unhandled exception :
[22:54:00 ERROR]: Paper Async Command Builder Thread Pool - 0
java.util.ConcurrentModificationException: null
        at java.util.TreeMap$PrivateEntryIterator.nextEntry(Unknown Source) ~[?:?]
        at java.util.TreeMap$ValueIterator.next(Unknown Source) ~[?:?]
        at net.minecraft.commands.Commands.fillUsableCommands(Commands.java:475) ~[?:?]
        at net.minecraft.commands.Commands.sendAsync(Commands.java:441) ~[?:?]
        at net.minecraft.commands.Commands.lambda$sendCommands$5(Commands.java:414) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) ~[?:?]
        at java.lang.Thread.run(Unknown Source) ~[?:?]
```

Unfortunately, I've been unable to reliably reproduce it once I started. Maybe it's a race condition? That'd be fun!

### What should have happened

Command `/stuff` should be unregistered without an exception.

### Server logs and CommandAPI config

_No response_

### Other

_No response_

willkroboth:
This PR fixes #494

https://github.com/PaperMC/Paper/pull/3116 added a [patch](https://github.com/PaperMC/Paper/blob/4b0bc74c90582f2d52d720c795228130545cd103/patches/server/0294-Async-command-map-building.patch) to Paper that causes the [Commands](https://wiki.vg/Protocol#Commands) packet to be built asynchronously. Specifically, they added the following [`ThreadPoolExector`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html) to the `net.minecraft.commands.Commands` class and used it to move some of the work off the main thread.

```java
public static final java.util.concurrent.ThreadPoolExecutor COMMAND_SENDING_POOL = new java.util.concurrent.ThreadPoolExecutor(
    0, 2, 60L, java.util.concurrent.TimeUnit.SECONDS,
    new java.util.concurrent.LinkedBlockingQueue<>(),
    new com.google.common.util.concurrent.ThreadFactoryBuilder()
        .setNameFormat("Paper Async Command Builder Thread Pool - %1$d")
        .setUncaughtExceptionHandler(new net.minecraft.DefaultUncaughtExceptionHandlerWithName(net.minecraft.server.MinecraftServer.LOGGER))
        .build(),
    new java.util.concurrent.ThreadPoolExecutor.DiscardPolicy()
);
```

Previously, if this thread pool happened to be looping through the Brigadier Vanilla or Resources CommandDispatcher and the CommandAPI registered or unregistered a command, a `ConcurrentModificationException` could occur.

This PR fixes this by adding a `modifyCommandTreesSafely` method to `PaperImplementations`. Anytime CommandAPIBukkit needs to modify the command trees, it calls this method. If the `COMMAND_SENDING_POOL` is present, the modification task will be submitted to the pool, and the current thread will be blocked until the task is completed. If the thread pool is not present, the task is run immediately like before.

 ---

I chose this solution because I noticed that the `COMMAND_SENDING_POOL` seemed to only run one task at a time. So, if you put the modify task into the pool, you can ensure that the pool isn't running another task, and no CME occurs. 

However, the fact that the pool only ran one task at a time struck me as odd. The javadocs for the `ThreadPoolExecutor` constructor indicate the second parameter — `2` — is the maximum number of threads allowed in the pool. Therefore, I expected that at most 2 tasks could be run at the same time in this pool, potentially allowing a modify task **and** a command build task to run at the same time, causing a CME.

In practice, I only saw this pool running at most 1 task. I don't really understand what's going on, so it seems possible that this PR could actually not solve the problem.

I think a better solution would be for `PaperImplementations#modifyCommandTreesSafely` to block the thread until the `COMMAND_SENDING_POOL` has finished all its tasks. However, I didn't see any way to do this without controlling how tasks are submitted to the pool.

Anyway, I am definitely not familiar with multithreaded code, so I have no idea if what I've done here is a 'good' solution. I'm not sure if I'm submitting tasks correctly or handling stuff like `InterruptedException` safely. Feedback is greatly appreciated.

---

So far, I've tested this manually on Paper and Spigot 1.20.2. It seems to resolve all the issues mentioned in #494. I plan to test this on other versions as well.

TODO:
- Test on real servers (Paper+Spigot)
  - [ ] 1.15, 1.15.1, 1.15.2
  - [ ] 1.16.1
  - [ ] 1.16.2, 1.16.3
  - [ ] 1.16.4
  - [ ] 1.16.5
  - [ ] 1.17
  - [ ] 1.17.1
  - [ ] 1.18, 1.18.1
  - [ ] 1.18.2
  - [ ] 1.19
  - [ ] 1.19.1, 1.19.2
  - [ ] 1.19.3
  - [ ] 1.19.4
  - [ ] 1.20, 1.20.1
  - [ ] 1.20.2

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperImplementations.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperImplementations.java
@@ -9,6 +9,13 @@ import org.bukkit.plugin.Plugin;
 
 import dev.jorel.commandapi.nms.NMS;
 import io.papermc.paper.event.server.ServerResourcesReloadedEvent;
+import org.jetbrains.annotations.Nullable;
+
+import java.lang.reflect.Field;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.function.Supplier;
 
 public class PaperImplementations {
 
@@ -16,6 +23,7 @@ public class PaperImplementations {
 	private final boolean isFoliaPresent;
 	private final NMS<?> nmsInstance;
 	private final Class<? extends CommandSender> feedbackForwardingCommandSender;
+	private final ThreadPoolExecutor paperCommandSendingPool;
 
 	/**
 	 * Constructs a PaperImplementations object
@@ -36,8 +44,24 @@ public class PaperImplementations {
 		} catch (ClassNotFoundException e) {
 			// uhh...
 		}
-		
 		this.feedbackForwardingCommandSender = tempFeedbackForwardingCommandSender;
+
+        this.paperCommandSendingPool = getThreadPoolExecutor(nmsInstance);
+	}
+
+	@Nullable
+	private static ThreadPoolExecutor getThreadPoolExecutor(NMS<?> nmsInstance) {
+		Class<?> nmsCommandsClass = nmsInstance.getNMSCommandsClass();
+		try {
+			// public static field
+			Field commandSendingPoolField = nmsCommandsClass.getField("COMMAND_SENDING_POOL");
+			return (ThreadPoolExecutor) commandSendingPoolField.get(null);
+		} catch (NoSuchFieldException | IllegalAccessException e) {
+			// Not a big deal, if the thread pool doesn't exist, then the server also isn't using it
+			//  We don't need to worry about ConcurrentModificationExceptions if Commands packets are
+			//  being built synchronously
+			return null;
+		}
 	}
 
 	/**
@@ -95,6 +119,61 @@ public class PaperImplementations {
 	public Class<? extends CommandSender> getFeedbackForwardingCommandSender() {
 		return this.feedbackForwardingCommandSender;
 	}
-	
 
+	/**
+	 * Runs a task that modifies the command trees using Paper's {@code COMMAND_SENDING_POOL}, if it is present. This
+	 * ensures that the command trees are not modified while Paper is building a Commands packet asynchronously, which
+	 * may cause a ConcurrentModificationException.
+	 * <p>
+	 * If the {@code COMMAND_SENDING_POOL} is not present (probably because we're on Spigot), the task is run
+	 * immediately, since there isn't anything building Commands packets async anyway.
+	 *
+	 * @param modifyTask The task to run that modifies the command trees.
+	 */
+	public void modifyCommandTreesSafely(Runnable modifyTask) {
+		try {
+			modifyCommandTreesSafely((Callable<Object>) () -> {
+				modifyTask.run();
+				return null;
+			});
+		} catch (Exception e) {
+			// `modifyTask` should not be throwing any checked exceptions anyway
+			//  since that would violate the signature of Runnable
+			throw new RuntimeException(e);
+		}
+	}
+
+	/**
+	 * Runs a task that modifies the command trees using Paper's {@code COMMAND_SENDING_POOL}, if it is present. This
+	 * ensures that the command trees are not modified while Paper is building a Commands packet asynchronously, which
+	 * may cause a ConcurrentModificationException.
+	 * <p>
+	 * If the {@code COMMAND_SENDING_POOL} is not present (probably because we're on Spigot), the task is run
+	 * immediately, since there isn't anything building Commands packets async anyway.
+	 *
+	 * @param modifyTask The task to run that modifies the command trees.
+	 * @return The result of running the {@code modifyTask}.
+	 * @param <T> The class of the object returned by the {@code modifyTask}.
+	 */
+	public <T> T modifyCommandTreesSafely(Supplier<T> modifyTask) {
+		try {
+			return modifyCommandTreesSafely((Callable<T>) modifyTask::get);
+		} catch (Exception e) {
+			// `modifyTask` should not be throwing any checked exceptions anyway
+			//  since that would violate the signature of Supplier
+			throw new RuntimeException(e);
+		}
+	}
+
+	private <T> T modifyCommandTreesSafely(Callable<T> modifyTask) throws Exception {
+		if(paperCommandSendingPool == null) {
+			// If the server isn't building Commands packets async (probably because we're on Spigot),
+			//  it is safe to run the task immediately
+			return modifyTask.call();
+		}
+		// Otherwise, submit the modify task to the pool.
+		//  The pool only runs one task at a time, so this ensures we don't modify
+		//  the commands while a command-building process is reading them
+		return paperCommandSendingPool.invokeAll(List.of(modifyTask)).get(0).get();
+	}
 }
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
@@ -467,48 +467,50 @@ public abstract class CommandAPIBukkit<Source> implements CommandAPIPlatform<Arg
 	@Override
 	public void postCommandRegistration(RegisteredCommand registeredCommand, LiteralCommandNode<Source> resultantNode, List<LiteralCommandNode<Source>> aliasNodes) {
 		if(!CommandAPI.canRegister()) {
-			// Usually, when registering commands during server startup, we can just put our commands into the
-			// `net.minecraft.server.MinecraftServer#vanillaCommandDispatcher` and leave it. As the server finishes setup,
-			// it and the CommandAPI do some extra stuff to make everything work, and we move on.
-			// So, if we want to register commands while the server is running, we need to do all that extra stuff, and
-			// that is what this code does.
-			// We could probably call all those methods to sync everything up, but in the spirit of avoiding side effects
-			// and avoiding doing things twice for existing commands, this is a distilled version of those methods.
+			paper.modifyCommandTreesSafely(() -> {
+				// Usually, when registering commands during server startup, we can just put our commands into the
+				// `net.minecraft.server.MinecraftServer#vanillaCommandDispatcher` and leave it. As the server finishes setup,
+				// it and the CommandAPI do some extra stuff to make everything work, and we move on.
+				// So, if we want to register commands while the server is running, we need to do all that extra stuff, and
+				// that is what this code does.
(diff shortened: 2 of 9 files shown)


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperImplementations.java
```diff
@@ -9,13 +9,21 @@
 
 import dev.jorel.commandapi.nms.NMS;
 import io.papermc.paper.event.server.ServerResourcesReloadedEvent;
+import org.jetbrains.annotations.Nullable;
```
JorelAli, line 12: Not a strict rule, but whenever possible I'd like to avoid using `Nullable` from `org.jetbrains.annotations`. There's a better solution that does (effectively) the same thing: `javax.annotation.Nullable`

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperImplementations.java
```diff
@@ -36,8 +44,24 @@ public PaperImplementations(boolean isPaperPresent, boolean isFoliaPresent, NMS<
...
+
+        this.paperCommandSendingPool = getThreadPoolExecutor(nmsInstance);
+	}
+
+	@Nullable
+	private static ThreadPoolExecutor getThreadPoolExecutor(NMS<?> nmsInstance) {
+		Class<?> nmsCommandsClass = nmsInstance.getNMSCommandsClass();
+		try {
+			// public static field
+			Field commandSendingPoolField = nmsCommandsClass.getField("COMMAND_SENDING_POOL");
```
JorelAli, line 57: This temporary variable is declared and then immediately used. We don't need to create this temporary variable here.

JorelAli:
_(For what it's worth, my multi-threaded programming knowledge is very limited, so I have no comments to make about the implementation from this PR)_

willkroboth:
Ah, I think I figured out why the pool only uses one thread. From the [javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html):

>  Queuing
Any [BlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/BlockingQueue.html) may be used to transfer and hold submitted tasks. The use of this queue interacts with pool sizing:
> - If fewer than corePoolSize threads are running, the Executor always prefers adding a new thread rather than queuing.
> - If corePoolSize or more threads are running, the Executor always prefers queuing a request rather than adding a new thread.
> - If a request cannot be queued, a new thread is created unless this would exceed maximumPoolSize, in which case, the task will be rejected.
>
> There are three general strategies for queuing:
> 1. Direct handoffs. A good default choice for a work queue is a [SynchronousQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/SynchronousQueue.html) that hands off tasks to threads without otherwise holding them. Here, an attempt to queue a task will fail if no threads are immediately available to run it, so a new thread will be constructed. This policy avoids lockups when handling sets of requests that might have internal dependencies. Direct handoffs generally require unbounded maximumPoolSizes to avoid rejection of new submitted tasks. This in turn admits the possibility of unbounded thread growth when commands continue to arrive on average faster than they can be processed.
> 2. Unbounded queues. Using an unbounded queue (for example a [LinkedBlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedBlockingQueue.html) without a predefined capacity) will cause new tasks to wait in the queue when all corePoolSize threads are busy. Thus, no more than corePoolSize threads will ever be created. (And the value of the maximumPoolSize therefore doesn't have any effect.) This may be appropriate when each task is completely independent of others, so tasks cannot affect each others execution; for example, in a web page server. While this style of queuing can be useful in smoothing out transient bursts of requests, it admits the possibility of unbounded work queue growth when commands continue to arrive on average faster than they can be processed.
> 3. Bounded queues. A bounded queue (for example, an [ArrayBlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ArrayBlockingQueue.html)) helps prevent resource exhaustion when used with finite maximumPoolSizes, but can be more difficult to tune and control. Queue sizes and maximum pool sizes may be traded off for each other: Using large queues and small pools minimizes CPU usage, OS resources, and context-switching overhead, but can lead to artificially low throughput. If tasks frequently block (for example if they are I/O bound), a system may be able to schedule time for more threads than you otherwise allow. Use of small queues generally requires larger pool sizes, which keeps CPUs busier but may encounter unacceptable scheduling overhead, which also decreases throughput.

So, looking at Paper's `ThreadPoolExecutor`:

```java
public static final java.util.concurrent.ThreadPoolExecutor COMMAND_SENDING_POOL = new java.util.concurrent.ThreadPoolExecutor(
    0, 2, 60L, java.util.concurrent.TimeUnit.SECONDS,
    new java.util.concurrent.LinkedBlockingQueue<>(),
    new com.google.common.util.concurrent.ThreadFactoryBuilder()
        .setNameFormat("Paper Async Command Builder Thread Pool - %1$d")
        .setUncaughtExceptionHandler(new net.minecraft.DefaultUncaughtExceptionHandlerWithName(net.minecraft.server.MinecraftServer.LOGGER))
        .build(),
    new java.util.concurrent.ThreadPoolExecutor.DiscardPolicy()
);
```

They're using an unbounded `LinkedBlockingQueue`, so that falls into number 2. As the javadocs explain, a thread pool will prefer to queue a task rather than create a new thread once `corePoolSize` threads have been created. Since the queue is unbounded, it can always queue a task and will never create more than `corePoolSize` threads, making `maximumPoolSize` irrelevant. Since `corePoolSize` is 0 here, once it makes its first thread, it has more threads than `corePoolSize`, so it won't make another thread. Hence, this thread pool will only use one thread.

---

So, that means this PR should work. Submitting the CommandAPI's register and unregister tasks to the thread pool will ensure that no CMEs happen. The pool only uses one thread at a time, so its either reading from the dispatchers to make a Commands packet or modifying the dispatchers as the CommandAPI requests.

However, this might be relying on unintended behavior. Given that `maximumPoolSize` is set to 2, Paper might have expected 2 threads to be handling Commands packet building. So, Paper might change this behavior in the future, breaking the changes here. That's just speculation though, it's hard to tell what Paper was thinking. https://github.com/PaperMC/Paper/pull/8170 created the `COMMAND_SENDING_POOL` and https://github.com/PaperMC/Paper/pull/8272 tweaked the configuration. Neither seem to give any reasoning why `maximumPoolSize` is set to 2. Intrestingingly, in some [discord messages](https://discord.com/channels/289587909051416579/925530366192779286/999453045056360528) discussing the changes, `maximumPoolSize` is set to 1. So, maybe the value of `2` is unintended ¯\\_(ツ)_/¯.

---

Anyway, the implementation for `PaperImplementations#modifyCommandTreesSafely` seems like it should work. I'm going to read through the javadocs for `ThreadPoolExecutor` some more to see if I can learn how I should be submitting tasks and handling threading exceptions.

willkroboth:
Hm, unfortunately, this fix does not solve the problem on all versions of Paper. The `COMMAND_SENDING_POOL` was only added by https://github.com/PaperMC/Paper/pull/8170 (build paper-1.19.2-117), so this PR only fixes the problem for builds after that PR. Earlier Paper builds can still have the CME (for example, see this [log](https://paste.gg/p/anonymous/debfa38ecf7f42c097b57d975145931d) on 1.19.2-116).

Specifically, the original PR (https://github.com/PaperMC/Paper/pull/3116, build paper-1.15.2-177) submits its task to the common `ForkJoinPool` (¯\\_(ツ)_/¯ whatever that is)
```java
java.util.concurrent.ForkJoinPool.commonPool().execute(() -> {
    sendAsync(entityplayer);
});
```

A random commit later on (https://github.com/PaperMC/Paper/commit/7323594d59206e41a39405e09fa5394c942fe5c6, build paper-1.18-31) changed this to use some sort of `MCUtil` task executor:
```java
net.minecraft.server.MCUtil.scheduleAsyncTask(() -> this.sendAsync(player));
```

For reference, the current system (https://github.com/PaperMC/Paper/pull/8170, build paper-1.19.2-117) changed it to this:
```java
COMMAND_SENDING_POOL.execute(() -> {
        this.sendAsync(player);
});
```

So, there needs to be more work to fix this on older Paper versions.

willkroboth:
Hmm, the current solution doesn't quite work. For versions ~1.15 to 1.18 `net.minecraft.server.CommandDispatcher` submits tasks using `ForkJoinPool#commonPool`, which references the field `ForkJoinPool.common`. To intercept the submitted tasks, I tried using reflection to replace that field with a CommandAPI-controlled object. Unfortunately, reflection is not allowed to access this field:

```java
java.lang.reflect.InaccessibleObjectException: 
  Unable to make field static final java.util.concurrent.ForkJoinPool 
    java.util.concurrent.ForkJoinPool.common accessible: 
  module java.base does not "opens java.util.concurrent" to unnamed module @14aca5ab
```

I'll need to find something else for this again

willkroboth:
On pause until #517. That PR splits up the Spigot and Paper modules, which will be helpful since this PR won't have to do separate Paper detection that works across all versions.

Issue #367, opened 2022-11-21
Allow changing namespace of registered commands


willkroboth:
### Description

This suggestion started as a question by Adixe [in discord](https://discord.com/channels/745416925924032523/745757702608912556/1044285197140627466)

Currently, whenever a command is registered, it is automatically registered under the `minecraft` namespace. So, if you register a command `/hello`, you can also run it using `/minecraft:hello`. (https://commandapi.jorel.dev/8.5.1/intro.html#how-the-commandapi-works)

It would be nice if developers could control what namespace their commands get registered under. 

### Expected code

```java
// default - runnable as /command or /minecraft:command
new CommandAPICommand("command").register();

// plugin namespace - runnable as /command or /pluginname:command
new CommandAPICommand("command").register(Plugin);

// custom namespace - runnable as /command or /custom:command
new CommandAPICommand("command").register("custom");

// no namespace - only runnable as /command
new CommandAPICommand("command").register("");
```

### Extra details

The `org.bukkit.SimpleCommandMap` class handles tracking, tab-completing, and dispatching all commands on the server, including CommandAPICommands via the VanillaCommandWrapper class.
```java
package org.bukkit.command;

public class SimpleCommandMap implements CommandMap {
    protected final Map<String, Command> knownCommands = new HashMap();

    // Pretty much what registering does
    public register(String name, String nameSpace, Command command) {
         this.knownCommands.put(nameSpace + ":" + name, command);
         this.knownCommands.put(name, command);
     }

    @Nullable
    public Command getCommand(@NotNull String name) {
        Command target = (Command)this.knownCommands.get(name.toLowerCase(Locale.ENGLISH));
        return target;
    }

    public boolean dispatch(@NotNull CommandSender sender, @NotNull String commandLine) throws CommandException {
        String[] args = commandLine.split(" ");
        if (args.length == 0) {
            return false;
        } else {
            String sentCommandLabel = args[0].toLowerCase(Locale.ENGLISH);
            Command target = this.getCommand(sentCommandLabel);
            if (target == null) {
                return false;
            } else {
                 // execute the target Command
            }
        }
    }

    public List<String> tabComplete(@NotNull CommandSender sender, @NotNull String cmdLine, @Nullable Location location) {
        int spaceIndex = cmdLine.indexOf(32);
        if (spaceIndex == -1) {
            ArrayList<String> completions = new ArrayList();
            Map<String, Command> knownCommands = this.knownCommands;
[...]

PR #509, opened 2023-11-30
Add support for custom namespaces for commands

DerEchtePilz:
This PR adds the ability to register commands with other namespaces than just `minecraft:`.
However, `minecraft:` is still the default if no namespace is given.

A bit of works still needs to be done:
- [ ] The Kotlin DSL needs to be updated to support this feature
- [ ] Documentation
- [ ] Velocity needs to be figured out
- [ ] Tests need to be figured out
- [ ] Global changelog entry
- [ ] Maybe a bit more, depending on what else comes to my mind or something that reviews suggest


commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -31,6 +31,7 @@
 import dev.jorel.commandapi.exceptions.GreedyArgumentException;
 import dev.jorel.commandapi.exceptions.MissingCommandExecutorException;
 import dev.jorel.commandapi.exceptions.OptionalArgumentException;
+import org.jetbrains.annotations.NotNull;
```
JorelAli, line 34: As opposed to `javax.annotation.Nonnull`? This removes the dependency on jetbrains which is unnecessary an excessive given Java already has one built-in.

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -272,13 +273,18 @@ private void checkHasExecutors() {
 
 	@Override
 	public void register() {
+		register("minecraft");
+	}
+
+	@Override
+	public void register(@NotNull String namespace) {
```
JorelAli, line 280: I don't tend to explicitly add `@Nonnull` annotations on parameters - our assumption is that everything that a user interacts with in the CommandAPI will be non-null _unless_ explicitly specified with a `@Nullable` annotation, suitable JavaDocs and something to bring a reader's attention in the documentation that it may be null. That way, if we ever get any null pointer exceptions, we can safely assume it's _always a user error_.

Otherwise, if you're going to add a not-null annotation explicitly, I'd expect a null pointer exception check as the first statement in this method.

commandapi-core/pom.xml
```diff
@@ -57,6 +57,11 @@
 			<version>2.19.0</version>
 			<scope>provided</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.jetbrains</groupId>
+			<artifactId>annotations</artifactId>
+			<version>24.0.1</version>
+		</dependency>
```
JorelAli, line 64: Unnecessary dependency, Java already has `javax.annotation.Nonnull`.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.15/src/main/java/dev/jorel/commandapi/nms/NMS_1_15.java
```diff
@@ -974,4 +976,17 @@ public <T> T getMinecraftServer() {
 		}
 	}
 
+	@Override
+	public void syncCommands() {
+		// For some reason, 1.15 is being stupid and has a private syncCommands method
+		// Reflection it is!
+		try {
+			Method syncCommands = Class.forName("org.bukkit.craftbukkit.v1_15_R1.CraftServer").getDeclaredMethod("syncCommands");
```
JorelAli, line 984: VarHandles please, safe reflection annotation needed as well.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.15/src/main/java/dev/jorel/commandapi/nms/NMS_1_15.java
```diff
@@ -974,4 +976,17 @@ public <T> T getMinecraftServer() {
...
+			syncCommands.setAccessible(true);
+			syncCommands.invoke(Bukkit.getServer());
+		} catch (ClassNotFoundException | NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
```
JorelAli, line 987: Did you know that `ReflectiveOperationException` covers all of these four exceptions?

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
DerEchtePilz: Oh right, that's what it was called. Thanks, will add. (I knew there was a native one, but typed `@Not` before checking suggestion)

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.15/src/main/java/dev/jorel/commandapi/nms/NMS_1_15.java
DerEchtePilz: No, I did not! That is very useful to know!

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.15/src/main/java/dev/jorel/commandapi/nms/NMS_1_15.java
DerEchtePilz: How would that look? Do VarHandles work with methods?

DerEchtePilz:
> It's also easy to remove those commands from the resources dispatcher ourselves, similar to what CommandAPIBukkit#unregisterInternal does. This way we also don't need to call syncCommands again.

In fact, I did what unregisterInternal does. I double checked any mistakes I may have made. It didn't work until I called syncCommands.
Although, if there is a way to get rid of the extra call to syncCommands, I would be really happy to remove it.
As for the current structure, I think the way it is set up currently makes the most sense. At least, it makes the most sense to me. Again, happy to take suggestions if it generally improves the readability of the code and/or removes some extra stuff I am doing right now which doesn't necessarily need to be there.

willkroboth:
> In fact, I did what unregisterInternal does. It didn't work until I called syncCommands.

Ah, I think the one detail missing from `setupNamespaces` compared to `unregisterInternal` is this line of code:

https://github.com/JorelAli/CommandAPI/blob/38a1a7682633a2095366095db183850e1a9d99eb/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L594-L598

... which uses these functions:

https://github.com/JorelAli/CommandAPI/blob/38a1a7682633a2095366095db183850e1a9d99eb/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L611-L650

The important thing happening there is that commands are being removed from what I call the "resources dispatcher". Bukkit copies commands from the vanilla `CommandDispatcher` into its own `CommandMap`, which is accessible in the API. Notably, the `CommandMap` is also used for tab-completing and running console commands. However, the code that sends commands to the player and handles player tab-complete and execution is based on Brigadier. So, Bukkit copies all the commands in the `CommandMap` into a second `CommandDispatcher`, the "resources dispatcher".

The `syncCommands` method is the method Bukkit calls to copy commands from the `CommandMap` to the "resources dispatcher". If you call it again, I'm pretty sure it does properly handle rebuilding the "resources dispatcher". However, it does process all of the commands from scratch. That's probably fine here since `setupNamespaces` is only called once, but I wanted to avoid it in `unregisterInternal` since that could be called many times.

If you do want to avoid calling `syncCommands`, you'll just need to add and remove same commands from the "resources dispatcher" as you're already doing for the `CommandMap`. My suggestion makes it so you only need to remove commands from the `CommandMap` and "resources dispatcher", but it should work either way. My suggestion also avoids the double nested loop over all known commands for all `RegisteredCommands` to find matches (and searching through the alias list), which could be a lot of work. It's probably fine since `setupNamespaces` is only called once, but it's just something to think about in combination with how Velocity namespaces will work.

willkroboth:
Yeah, currently, Velocity commands do not have namespaces. Since it's just a proxy, it doesn't include Minecraft server commands like Bukkit does, so I guess they didn't bother with automatically assigning `minecraft` or plugin namespaces. We can definitely still create namespaced versions of commands ourself if we want to (just put register `namespace:command` as well), and I think it is helpful in case another plugin tries to use the same command name.

Velocity plugins do have IDs (see https://docs.papermc.io/velocity/dev/api-basics#create-the-plugin-class), which would make sense as the namespace to use for a Velocity Plugin. It looks like you can get that ID given a plugin object with something like this:

```java
CommandAPIVelocity.getConfiguration().getServer().getPluginManager().fromInstance(/* Plugin object here */).get().getDescription().getId();
```

DerEchtePilz:
So this is basically done. I need to test compatibility with #417 because when I read it I got worried it would only work with a `minecraft:` namespace. It's probably fine, I just want to make sure.
As for the tests, I tested this ingame extensively so with confidence I can say it works. Also, I have no idea how to test for a correct namespace registration so I'll leave this to someone who knows more :D

@willkroboth Would be great if you'd review this, after all you made this issue #367 and probably had an idea how it could work.

willkroboth (changes requested):

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -270,15 +272,32 @@ private void checkHasExecutors() {
 		}
 	}
 
+	/**
+	 * Registers the command with the minecraft: namespace
+	 */
 	@Override
 	public void register() {
+		register("minecraft");
```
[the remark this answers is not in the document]
DerEchtePilz: Really good idea, didn't think about that!

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.16.1/src/main/java/dev/jorel/commandapi/nms/NMS_1_16_R1.java
```diff
@@ -1050,5 +1050,4 @@ public <T> T getMinecraftServer() {
 			return null;
 		}
 	}
-
```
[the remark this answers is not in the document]
DerEchtePilz: Those lines are probably a relic from when I removed the syncCommands method. I didn't bother to remove the NMS class diffs but it is probably a good idea to not make them show up here.

commandapi-core/src/main/java/dev/jorel/commandapi/AbstractCommandAPICommand.java
```diff
@@ -270,15 +272,32 @@ private void checkHasExecutors() {
...
+	}
+
+	/**
+	 * Registers the command with a given namespace
+	 *
+	 * @param namespace The namespace of this command. This cannot be null
+	 * @throws NullPointerException if the namespace is null
+	 */
+	@Override
+	public void register(String namespace) {
```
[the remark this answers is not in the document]
willkroboth: I don't think command conflicts with no namespace is too big of a problem. Currently, when registering two commands via the CommandAPI with the same name they conflict. Even though they get the default `minecraft` namespace, that's the same namespace and they conflict anyway. They would need to be explicitly given different names or namespaces to solve the problem anyway.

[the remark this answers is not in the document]
DerEchtePilz: Fair.
I personally wouldn't implement it but I guess you really thought about it so I guess we could implement that.

commandapi-kotlin/commandapi-velocity-kotlin/src/main/kotlin/dev/jorel/commandapi/kotlindsl/CommandAPICommandDSL.kt
```diff
@@ -8,6 +8,7 @@ import dev.jorel.commandapi.arguments.*
 import java.util.function.Predicate
 
 inline fun commandAPICommand(name: String, command: CommandAPICommand.() -> Unit = {}) = CommandAPICommand(name).apply(command).register()
+inline fun commandAPICommand(name: String, namespace: String, command: CommandAPICommand.() -> Unit = {}) = CommandAPICommand(name).apply(command).register(namespace)
```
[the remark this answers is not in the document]
DerEchtePilz: Yeah, Bukkit has something similar but I didn't deal with it because I don't think it's possible with the way we set up commands.
If you feel like there's a way, feel free to contribute to this branch. Obviously, associating commands with plugins should only happen when the `register` method with the plugin object is used - if it's possible at all.

DerEchtePilz:
So, this is done. If there's nothing else, we could even say we still put this into 9.3.0.

commandapi-platforms/commandapi-velocity/commandapi-velocity-core/src/main/java/dev/jorel/commandapi/CommandAPIVelocity.java
```diff
@@ -215,8 +215,18 @@ public void postCommandRegistration(RegisteredCommand registeredCommand, Literal
...
+		LiteralCommandNode<CommandSource> builtNode = getBrigadierDispatcher().register(node);
+		if (namespace.isEmpty()) {
+			// We're done. The command already is registered
+			return builtNode;
+		}
+		getBrigadierDispatcher().register(LiteralArgumentBuilder.<CommandSource>literal(namespace + ":" + node.getLiteral())
+			.redirect(builtNode)
+			.requires(node.getRequirement())
+			.executes(builtNode.getCommand())
+		);
```
willkroboth, line 228: I was thinking about why `CraftServer#syncCommands` on Bukkit copied the command nodes (see [`CommandAPIBukkit#namespaceNode`](https://github.com/JorelAli/CommandAPI/blob/0f8dd2d05e88c168fc1fd7c8894a234ed2cc1fd3/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L528-L543)) rather than using a redirect like we have here. I don't know if they had this problem, but I realized that we probably should be copying the nodes. Consider registering these commands:

```java
new CommandAPICommand("command")
	.withArguments(new LiteralArgument("a"))
	.executes(/* executorA */)
	.register("a");

new CommandAPICommand("command")
	.withArguments(new LiteralArgument("b"))
	.executes(/* executorB */)
	.register("b");
```

Right now, this would create the following command structure:

```
literal `command`:
    literal `a`:
        executes `executorA`
    literal `b`:
        executes `executorB`

literal `a:command`:
    redirects to literal `command`
literal `b:command`:
    redirects to literal `command`
```

The two commands get merged into the `command` literal, which is a little weird, but I don't think that can be avoided. On the other hand, `a:command b` becomes a valid command, even though `a:command` is only supposed to have the `a` literal available. Since the namespaced nodes redirect to the original node, they also get merged together. However, we can avoid this merging by copying the nodes (something similar would need to be implemented in `CommandAPIBukkit` as well), like so:

```java
@Override
public LiteralCommandNode<CommandSource> registerCommandNode(LiteralArgumentBuilder<CommandSource> node, String namespace) {
	LiteralCommandNode<CommandSource> builtNode = getBrigadierDispatcher().register(node);
	if (!namespace.isEmpty()) {
		getBrigadierDispatcher().addChild(namespaceNode(builtNode, namespace));
	}
	return builtNode;
}

private LiteralCommandNode<CommandSource> namespaceNode(LiteralCommandNode<CommandSource> original, String namespace) {
	// Adapted from a section of `org.bukkit.craftbukkit.CraftServer#syncCommands`
	LiteralCommandNode<CommandSource> clone = new LiteralCommandNode<>(
		namespace + ":" + original.getLiteral(),
		original.getCommand(),
		original.getRequirement(),
		original.getRedirect(),
		original.getRedirectModifier(),
		original.isFork()
	);

	for (CommandNode<CommandSource> child : original.getChildren()) {
		clone.addChild(child);
	}

	return clone;
}
```

Then, the example commands from before would look like this:

```
literal `command`:
    literal `a`:
        executes `executorA`
    literal `b`:
        executes `executorB`

literal `a:command`:
    literal `a`:
        executes `executorA`
literal `b:command`:
    literal `b`:
        executes `executorB`
```

The un-namespaced version of the commands still get messed up, but giving the two commands different namespaces allows a namespaced version of the commands to be used as intended without conflict.

As a side note, `CommandAPIBukkit#namespaceNode` doesn't do anything Bukkit-specific; it just needs Brigadier. If we do implement the node-copying solution over here in Velocity, it might make sense to move that method to `CommandAPIHandler` so the logic can be shared.

willkroboth (changes requested):
We should definitely add some tests before merging. All the ways of setting namespaces should work. Two commands registered with different namespaces also shouldn't conflict. 

Everything should also work the same once these lines are called to simulate the server finishing enabling:

```java
disablePaperImplementations();
assertDoesNotThrow(() -> server.getScheduler().performOneTick());
```

We should make sure the namespace logic that is triggered after the server is enabled works the same as before enabling. The `minecraft:` version of commands should also be properly removed from the `CommandMap` and resources dispatcher.

Those tests can probably be added to [`CommandRegistrationTests`](https://github.com/JorelAli/CommandAPI/blob/0f8dd2d05e88c168fc1fd7c8894a234ed2cc1fd3/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/CommandRegistrationTests.java) and [`OnEnableTests`](https://github.com/JorelAli/CommandAPI/blob/0f8dd2d05e88c168fc1fd7c8894a234ed2cc1fd3/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/OnEnableTests.java). I'd be happy to figure out the `OnEnableTests`, mostly because they're a big mess that I should probably clean up anyway.

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -615,15 +614,16 @@ void register(CommandMetaData<CommandSender> meta, final Argument[] args,
...
+			fullDescription, usageDescription, aliases, permission, namespace);
 		registeredCommands.add(registeredCommandInformation);
 
 		// Handle previewable arguments
 		handlePreviewableArguments(commandName, args, aliases);
 
 		platform.preCommandRegistration(commandName);
 
-		CommandAPI.logInfo("Registering command /" + commandName + " " + humanReadableCommandArgSyntax);
+		String namespacedCommandName = (!namespace.isEmpty()) ? namespace + ":" + commandName : commandName;
```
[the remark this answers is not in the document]
DerEchtePilz: To be fair, what's there currently definitely is what I thought of first, I do think removing the negation is more readable though.
I will change that.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
```diff
@@ -530,8 +543,31 @@ private LiteralCommandNode<Source> namespaceNode(LiteralCommandNode<Source> orig
...
 
 	@Override
-	public LiteralCommandNode<Source> registerCommandNode(LiteralArgumentBuilder<Source> node) {
-		return getBrigadierDispatcher().register(node);
+	public LiteralCommandNode<Source> registerCommandNode(LiteralArgumentBuilder<Source> node, String namespace) {
+		LiteralCommandNode<Source> builtNode = getBrigadierDispatcher().register(node);
+		if(!namespace.equals("minecraft")) {
+			String name = node.getLiteral();
+			String namespacedName = (!namespace.isEmpty()) ? namespace + ":" + name : name;
+			if (name.equals(namespacedName)) {
```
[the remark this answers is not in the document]
DerEchtePilz: Yeah, that's a good idea, I will remove the `else if` though because the `builtNode` can just be returned instantly there.

DerEchtePilz:
Cool, apparently I committed something stupid and now the tests fail. I am not gonna do this today though.

commandapi-platforms/commandapi-velocity/commandapi-velocity-core/src/main/java/dev/jorel/commandapi/CommandAPIVelocity.java
DerEchtePilz: I don't really know what you mean here. Feel free to implement this yourself, I am a little overwhelmed here.

DerEchtePilz:
> We should definitely add some tests before merging. All the ways of setting namespaces should work. Two commands registered with different namespaces also shouldn't conflict.
> 
> Everything should also work the same once these lines are called to simulate the server finishing enabling:
> 
> ```java
> disablePaperImplementations();
> assertDoesNotThrow(() -> server.getScheduler().performOneTick());
> ```
> 
> We should make sure the namespace logic that is triggered after the server is enabled works the same as before enabling. The `minecraft:` version of commands should also be properly removed from the `CommandMap` and resources dispatcher.
> 
> Those tests can probably be added to [`CommandRegistrationTests`](https://github.com/JorelAli/CommandAPI/blob/0f8dd2d05e88c168fc1fd7c8894a234ed2cc1fd3/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/CommandRegistrationTests.java) and [`OnEnableTests`](https://github.com/JorelAli/CommandAPI/blob/0f8dd2d05e88c168fc1fd7c8894a234ed2cc1fd3/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/OnEnableTests.java). I'd be happy to figure out the `OnEnableTests`, mostly because they're a big mess that I should probably clean up anyway.

Yeah, well, I thought about tests too. It's probably enough to simply test the registration and then whether the correct values have been removed and added by simply invoking the commands.
It should be that easy, right?

commandapi-platforms/commandapi-velocity/commandapi-velocity-core/src/main/java/dev/jorel/commandapi/CommandAPIVelocity.java
DerEchtePilz: When writing tests, I registered this command, following your example command at the top:
```java
CommandAPICommand a = new CommandAPICommand("test")
	.withArguments(LiteralArgument.of("forced"))
	.executesPlayer(info -> {
		results.set("forced");
	});

CommandAPICommand b = new CommandAPICommand("test")
	.withArguments(LiteralArgument.of("notforced"))
	.executesPlayer(info -> {
		results.set("notforced");
	});

a.register("a");
b.register("b");
```

Then, I added tests that should succeed:
```java
server.dispatchCommand(player, "test forced");
assertEquals("forced", results.get());

server.dispatchCommand(player, "a:test forced");
assertEquals("forced", results.get());

server.dispatchCommand(player ,"test notforced");
assertEquals("notforced", results.get());

server.dispatchCommand(player, "b:test notforced");
assertEquals("notforced", results.get());
```

Last, I added tests that should fail:
```java
assertThrows(CommandSyntaxException.class, () -> server.dispatchThrowableCommand(player, "a:test notforced"));
assertThrows(CommandSyntaxException.class, () -> server.dispatchThrowableCommand(player, "b:test forced"));
```
This also succeeded. So I am not really sure if this is a problem _somewhere_ but redirecting seems to work absolutely fine.

willkroboth: Ah, I think that happens because the current test is being run after the server is enabled.

For example, this test succeeded:

<details>
<summary>Sucessful test</summary>

```java
@Test
public void registerCommandsWithDifferentNamespacesButSameName() {
	Mut<String> results = Mut.of();
	PlayerMock player = server.addPlayer();

	// Enabling the server before registering commands makes the test succeed
	disablePaperImplementations();
	Bukkit.getPluginManager().callEvent(new ServerLoadEvent(ServerLoadEvent.LoadType.STARTUP));
	assertDoesNotThrow(() -> server.getScheduler().performOneTick());
	assertFalse(CommandAPI.canRegister());

	// Test same command name and argument setup but different namespace
	CommandTree a = new CommandTree("test")
		.then(LiteralArgument.of("a").executesPlayer(info -> {
			results.set("a");
		}));

	CommandTree b = new CommandTree("test")
		.then(LiteralArgument.of("b").executesPlayer(info -> {
			results.set("b");
		}));

	a.register("a");
	b.register("b");

	// No namespace command should have been merged
	server.dispatchCommand(player, "test a");
	assertEquals("a", results.get());

	server.dispatchCommand(player, "test b");
	assertEquals("b", results.get());

	// a:test command should be separate from b branch
	server.dispatchCommand(player, "a:test a");
	assertEquals("a", results.get());

	assertThrows(CommandSyntaxException.class, () -> server.dispatchThrowableCommand(player, "a:test b"));

	// b:test command should be separated from a branch
	assertThrows(CommandSyntaxException.class, () -> server.dispatchThrowableCommand(player, "b:test a"));

	server.dispatchCommand(player, "b:test b");
	assertEquals("b", results.get());


	assertNoMoreResults(results);
}
```

</details>

However, this test fails:

<details>
<summary>Failing test</summary>

```java
/* CommandRegistrationTests.registerCommandsWithDifferentNamespacesButSameName:755 
    Expected com.mojang.brigadier.exceptions.CommandSyntaxException to be thrown, 
    but nothing was thrown. */

@Test
public void registerCommandsWithDifferentNamespacesButSameName() {
	Mut<String> results = Mut.of();
	PlayerMock player = server.addPlayer();

	// Test same command name and argument setup but different namespace
	CommandTree a = new CommandTree("test")
		.then(LiteralArgument.of("a").executesPlayer(info -> {
			results.set("a");
		}));

	CommandTree b = new CommandTree("test")
		.then(LiteralArgument.of("b").executesPlayer(info -> {
			results.set("b");
		}));

	a.register("a");
	b.register("b");

	// Enable server after registering commands to call "fixNamespaces"
	disablePaperImplementations();
	Bukkit.getPluginManager().callEvent(new ServerLoadEvent(ServerLoadEvent.LoadType.STARTUP));
	assertDoesNotThrow(() -> server.getScheduler().performOneTick());
	assertFalse(CommandAPI.canRegister());

	// No namespace command should have been merged
	server.dispatchCommand(player, "test a");
	assertEquals("a", results.get());

	server.dispatchCommand(player, "test b");
	assertEquals("b", results.get());

	// a:test command should be separate from b branch
	server.dispatchCommand(player, "a:test a");
	assertEquals("a", results.get());

	// This line (755) fails
	assertThrows(CommandSyntaxException.class, () -> server.dispatchThrowableCommand(player, "a:test b"));

	// b:test command should be separated from a branch
	assertThrows(CommandSyntaxException.class, () -> server.dispatchThrowableCommand(player, "b:test a"));

	server.dispatchCommand(player, "b:test b");
	assertEquals("b", results.get());


	assertNoMoreResults(results);
}
```

</details>

Whether or not the server is enabled matters because the two cases are handled by different code.

When the server is enabled, this code creates the namespaced version:

https://github.com/JorelAli/CommandAPI/blob/1952d332627ee2c04854c41b19997746c6e5cf3e/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L480-L549

This path uses `CommandAPIBukkit#namespaceNode` to copy the original node as a namespaced version.

When the server isn't enabled, it uses this code to create the namespaced version:

https://github.com/JorelAli/CommandAPI/blob/1952d332627ee2c04854c41b19997746c6e5cf3e/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L551-L572

This uses the redirect method. 

So, when the server is enabled, `a:test` is a clone of the `a` command, and it correctly can not access the `b` branch as you saw. However, if the server is not enabled, `a:test` is redirected to `test`, so running `a:test b` incorrectly succeeds, and the test fails. To solve this, we should probably always use `CommandAPIBukkit#namespaceNode` to copy the node instead of redirecting, like so:

```java
@Override
public LiteralCommandNode<Source> registerCommandNode(LiteralArgumentBuilder<Source> node, String namespace) {
	LiteralCommandNode<Source> builtNode = getBrigadierDispatcher().register(node);
	if (namespace.isEmpty()) {
		// Bukkit will automatically create `minecraft:command`
		// We want to remove that so there is no namespace
		namespacesToFix.add("minecraft:" + node.getLiteral());
	} else if(!namespace.equals("minecraft")) {
		// If Bukkit won't create our namespaces, we'll do it ourselves
		getBrigadierDispatcher().getRoot().addChild(namespaceNode(builtNode, namespace));

		// Make sure Bukkit's automatic `minecraft:` namespaces don't stay
		String name = node.getLiteral();
		String namespacedName = namespace + ":" + name;
		namespacesToFix.add("minecraft:" + name);
		namespacesToFix.add("minecraft:" + namespacedName);
	}
	return builtNode;
}
```

In general, since there are two separate code paths in action, we should probably ensure all the tests work the same when registering the commands before and after the server is enabled to catch cases like this. I also noticed that `CommandRegistrationTests` is pretty big, so it might be a good idea to separate the namespace tests into separate methods in another class called `CommandNamespaceTests` or something.

DerEchtePilz:
Alright. I now consider this feature finished. After multiple revisions started by @willkroboth pointing out mistakes I made, I am now confident in saying that this feature works now.

The current caveat is that some test don't succeed but manual test have confirmed that the failing automated tests actually succeed in a real Minecraft environment.

@willkroboth @JorelAli 
If one of you or even both of you could do a code review, I believe this PR could be merged until Sunday or so :)

willkroboth (changes requested):
No major changes, just some details I noticed in the implementation and wording of various bits and pieces.

I would prefer to resolve all the disabled tests before merging. While it is probably just a problem with the testing framework and doesn't affect a real server, leaving a disabled test without an explanation indicates that we don't understand why the code we wrote is doing what it does.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkitConfig.java
```diff
@@ -50,6 +51,19 @@ public CommandAPIBukkitConfig useMojangMappings(boolean useMojangMappings) {
...
+	/**
+	 * Sets whether the CommandAPI should use the plugin's name as the default namespace namespace
+	 * <p>
+	 * If called, any call to {@link CommandAPIConfig#setNamespace(String)} will be ignored
+	 *
+	 * @return this CommandAPIBukkitConfig
+	 */
+	public CommandAPIBukkitConfig usePluginNamespace() {
+		super.setNamespace(plugin.getName().toLowerCase());
+		super.usePluginNamespace = true;
```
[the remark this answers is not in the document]
DerEchtePilz: I was thinking about that method like a "this method should take highest priority" kind of thing. If you call that method I think it's safe to assume that people probably won't use `setNamespace(String)` after using `usePluginNamespace()` but I added that boolean just in case.
That's why the JavaDocs mention that any call to `CommandAPIConfig#setNamespace(String)` is ignored if `usePluginNamespace` is called.

So, in short: I was thinking of this method as highest priority but if we think that should not be the case, I am happy to remove it.

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -615,15 +618,22 @@ void register(CommandMetaData<CommandSender> meta, final Argument[] args,
...
 		RegisteredCommand registeredCommandInformation = new RegisteredCommand(commandName, argumentsString, shortDescription,
-			fullDescription, usageDescription, aliases, permission);
+			fullDescription, usageDescription, aliases, permission, namespace);
+		registeredCommands.add(registeredCommandInformation);
+
+		List<RegisteredCommand> registeredCommands = registeredCommandMap.containsKey(commandName)
+			? registeredCommandMap.get(commandName)
+			: new ArrayList<>();
 		registeredCommands.add(registeredCommandInformation);
+		registeredCommandMap.put(commandName, registeredCommands);
```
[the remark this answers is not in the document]
DerEchtePilz: Yup, will change 👍

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/CommandNamespaceTests.java
```diff
@@ -0,0 +1,767 @@
...
+		Mockito.when(null).thenReturn(handle);
+
+		return runCommandsPlayer;
+	}
+
+	/*********
+	 * Tests *
+	 *********/
+
+	@Disabled
```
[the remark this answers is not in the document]
DerEchtePilz: Please, go ahead. They do work in-game, I did verify that so that's nothing that should stop us from merging.
Figuring this out would be great though!

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/CommandNamespaceTests.java
```diff
@@ -0,0 +1,767 @@
...
+		assertCommandFailsWith(
+			player,
+			"minecraft:test b 123",
+			"Unknown or incomplete command, see below for error at position 0: <--[HERE]"
+		);
+
+		assertNoMoreResults(results);
+	}
+
+	@Disabled
```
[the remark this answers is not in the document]
DerEchtePilz: Please, go ahead. They do work in-game, I did verify that so that's nothing that should stop us from merging.
Figuring this out would be great though!

commandapi-platforms/commandapi-velocity/commandapi-velocity-core/src/main/java/dev/jorel/commandapi/CommandAPIVelocity.java
```diff
@@ -215,8 +215,31 @@ public void postCommandRegistration(RegisteredCommand registeredCommand, Literal
...
+			original.getRedirect(),
+			original.getRedirectModifier(),
+			original.isFork()
+		);
+
+		for (CommandNode<CommandSource> child : original.getChildren()) {
+			clone.addChild(child);
+		}
+		return clone;
 	}
```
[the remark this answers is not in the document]
DerEchtePilz: Yeah, makes sense!

docssrc/src/commandregistration.md
```diff
@@ -170,7 +177,19 @@ Executes a command regardless of what the command sender is, using the `NativePr
 void register()
 ```
 
-Registers the command.
+Registers the command with the default `minecraft` namespace.
```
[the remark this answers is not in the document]
DerEchtePilz: Yup, will change 👍

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/CommandNamespaceTests.java
```diff
@@ -0,0 +1,767 @@
...
+		assertEquals("b", results.get());
+		assertEquals("123", results.get());
+
+		server.dispatchCommand(player, "minecraft:test b 123");
+		assertEquals("b", results.get());
+		assertEquals("123", results.get());
+
+		CommandAPI.unregister("test", true);
+
+		command.register("namespace");
```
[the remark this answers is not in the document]
DerEchtePilz: Yeah, I got a bit lazy when writing these :)
Thought to myself that they get converted to `CommandAPICommand`s anyway so it's probably fine. As you've pointed out in another comment, there are some CommandTree-specific things to test so updating these would be wise.
Will change 👍

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkitConfig.java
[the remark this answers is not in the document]
JorelAli: > I don't think the `CommandAPIConfig#usePluginNamespace` field is necessary.

It isn't, but it's a convenience function. We can assume for all intents and purposes that a developer is not going to use `setNamespace()` twice and/or use `usePluginNamespace()` and `setNamespace` together.

It's probably "safer" to simply prevent calling it twice in the first place, but it's an extra safety measure that we could probably ignore. If a developer is complaining about namespaces acting funky, I'm more than happy to tell them to stop writing conflicting code.

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -615,15 +618,22 @@ void register(CommandMetaData<CommandSender> meta, final Argument[] args,
...
 
 		// Handle previewable arguments
 		handlePreviewableArguments(commandName, args, aliases);
 
 		platform.preCommandRegistration(commandName);
 
-		CommandAPI.logInfo("Registering command /" + commandName + " " + humanReadableCommandArgSyntax);
+		String namespacedCommandName = namespace.isEmpty() ? commandName : namespace +  ":" + commandName;
```
JorelAli, line 631: It's possible I've missed something, but is the only case when `namespace.isEmpty()` will occur be for Velocity?

DerEchtePilz: Nope, there's one instance in `CommandAPIVelocity#registerCommandNode(LiteralCommandNode, String)`

JorelAli: > Nope, there's one instance in `CommandAPIVelocity#registerCommandNode(LiteralCommandNode, String)`

In other words, will `namespace.isEmpty()` always be `false` when running on a Bukkit server?

DerEchtePilz: Actually, no. It is perfectly valid to register a command using a `""` String literal in the register(String) method

JorelAli: Should we be allowing a user to use `""` as a namespace in the `register(String)` method for Bukkit?

DerEchtePilz: I actually don't know and had the same concern. @willkroboth [mentioned](https://github.com/JorelAli/CommandAPI/pull/509#discussion_r1418932864) something about that here which is why I allowed that.

JorelAli: My counter-example to "a command with no namespace" is simply the fact that in the Bukkit command system, _all commands should have namespaces_. If I have two commands `minecraft:reload` and `bukkit:reload`, if I run `/reload`, it will choose the command with the highest "priority" according to however Bukkit assigns priority. In Bukkit's ecosystem, commands with "no namespace" are automatically given the namespace of their registering plugin.

From a Bukkit point of view, I don't think we should be allowing commands with no namespaces. While _yes_ they will probably all work out fine, semantically it's getting a little bit non-Bukkit-ty and I don't like the idea of that.

DerEchtePilz: I do agree with that. I absolutely do.
Another point I was thinking about were the currently existing exceptions. I originally added those really early when I still had the `@Nonnull` annotation on parameters.
I believe we could change this behaviour to instead log a message, maybe only if verbose outputs are true or something and register the command under the default namespace instead.
Or do we keep *an* exception that doesn't necessarily is a NPE.
Depending on this change we should be implementing a similar behaviour when an empty String is passed into that method.

willkroboth: > If I have two commands minecraft:reload and bukkit:reload, if I run /reload, it will choose the command with the highest "priority" according to however Bukkit assigns priority.

Yeah, that makes sense. A note on the priority system: Bukkit chooses whichever command is registered first. I think it would be good if we complied with this on the following line, by only adding our command to the map if there isn't already a key associated with the command name.

https://github.com/JorelAli/CommandAPI/blob/d37d424072d6381639d2027f2ab32bcabd0464aa/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L561

Imo, since Velocity has no-namespace commands as the default behavior, it makes sense for the API design to allow that in Bukkit as well. I think it's important to allow the developer to choose what they want to do. `CommandAPIHandler` can internally already handle an empty namespace for Velocity, so we can also just expose that to the developer in Bukkit as well.

DerEchtePilz: > Imo, since Velocity has no-namespace commands as the default behavior, it makes sense for the API design to allow that in Bukkit as well.

I am not so sure I agree with this. No-namespace commands are a Velocity feature already, on Bukkit they are not. We should follow this system - even if we could do different.
For me this means that we could allow no-namespace commands for Velocity since they are already a thing but we don't allow no-namespace commands on Bukkit because they are not a thing there.

willkroboth: 🤷 Alright, I suppose it helps simplify some logic.

For example, in these lines:

https://github.com/JorelAli/CommandAPI/blob/d37d424072d6381639d2027f2ab32bcabd0464aa/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L557-L574

Instead of using reflection to access the `knownCommands` map, we can just use [`CommandMap#register(String fallbackPrefix, Command command)`](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/command/CommandMap.html#register(java.lang.String,org.bukkit.command.Command)) since the namespace will no longer be an empty String.

DerEchtePilz: Do we want to cancel the command registration with an exception or a message here?
If a message, do we want to log that even if verbose outputs are false?

JorelAli: Let's use a message. We would probably want this to log despite verbose outputs set to false.

If the namespace is null, we should NPE.

If the namespace is empty, could we fall back to the `minecraft` for Bukkit (and if so, add that to the message as well?)

DerEchtePilz: Yeah, we would definitely fall back to the default namespace. I'd also be inclined to not throw a NPE and instead just log this as well and register using the default namespace. What do we think abou that?

JorelAli: There should never be a case where a user registers a command with a "null" namespace - a user should not be passing null to CommandAPI methods. It is probably worth adding an explicit `@Nonnull` annotation in this case. If a user wants "null" as a namespace, they can just use the method that doesn't accept a namespace.

DerEchtePilz: So? I am not sure how to continue now:
- Throwing a NPE if a namespace was unexpectedly null
- Logging a message if a namespace was unexpectedly null
- Adding a `@Nonnull` annotation and still do one of the methods above because that annotation doesn't prevent using null, but instead only displays a warning

JorelAli: If `null`, we throw an NPE (we don't log a message, we just throw). The only possible way for the namespace to be `null` is because a user has passed `null` to the namespace setting function, and we don't support that because users shouldn't be passing `null` to the CommandAPI.

We _could_ also add `@Nonnull`, but I don't think it's strictly necessary.

commandapi-platforms/commandapi-velocity/commandapi-velocity-core/src/main/java/dev/jorel/commandapi/CommandAPIVelocity.java
```diff
@@ -215,8 +215,14 @@ public void postCommandRegistration(RegisteredCommand registeredCommand, Literal
...
-	public LiteralCommandNode<CommandSource> registerCommandNode(LiteralArgumentBuilder<CommandSource> node) {
-		return getBrigadierDispatcher().register(node);
+	public LiteralCommandNode<CommandSource> registerCommandNode(LiteralArgumentBuilder<CommandSource> node, String namespace) {
+		LiteralCommandNode<CommandSource> builtNode = getBrigadierDispatcher().register(node);
+		if (namespace.isEmpty()) {
+			// We're done. The command already is registered
+			return builtNode;
+		}
+		getBrigadierDispatcher().getRoot().addChild(CommandAPIHandler.getInstance().namespaceNode(builtNode, namespace));
+		return builtNode;
```
JorelAli, line 225: Nitpick: This is a weird way to write this code. You have two `return builtNode` statements, but you don't ever change `builtNode` after the `if` statement.

You have this:
```java
if (statement) {
  return value;
}
doStuff()
return value;
```

Which is equivalent to this:
```java
if (statement) {
  return value;
} else {
  doStuff()
  return value;
}
```

Which is equivalent to this:

```java
if (!statement) {
  doStuff()
}
return value;
```

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -615,15 +619,17 @@ void register(CommandMetaData<CommandSender> meta, final Argument[] args,
 			argumentsString.add(arg.getNodeName() + ":" + arg.getClass().getSimpleName());
 		}
 		RegisteredCommand registeredCommandInformation = new RegisteredCommand(commandName, argumentsString, shortDescription,
-			fullDescription, usageDescription, aliases, permission);
+			fullDescription, usageDescription, aliases, permission, namespace);
 		registeredCommands.add(registeredCommandInformation);
+		registeredCommandMap.computeIfAbsent(commandName, k -> new ArrayList<>()).add(registeredCommandInformation);
```
JorelAli, line 624: My understanding is that `registeredCommandMap` now contains everything that `registeredCommands` contains. Is it possible that we can remove `registeredCommands` now by instead iterating over all lists of values in `registeredCommandMap` so we're not storing `[everything]` and `Map<[commands, [everything]]>`

DerEchtePilz: Probably. I haven't considered it. I kept `registeredCommands` because it makes `CommandAPI#getRegisteredCommands()` not change at all. I suppose we could be computing the return value there everytime that method is called. Obviously caching is also an option but considering we support dynamically registering commands I am not sure this is a viable option.
Additionally, I only added the `registeredCommandMap` so `CommandAPIBukkit#fixPermissions()` doesn't have to loop over the whole list everytime a permission is "fixed".

willkroboth: While figuring out why the permissions tests aren't working I modified the code so that `registeredCommandMap` is no longer needed. I haven't fully gotten the tests working yet though, so I'm not entirely sure that it works. However, I think I may have found some inconsistent behavior with the current solution, but again I need to test that first.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPICommand.java
```diff
@@ -22,4 +23,27 @@ protected CommandAPICommand newConcreteCommandAPICommand(CommandMetaData<Command
...
+	 * Registers the command with a given namespace
+	 *
+	 * @param namespace The namespace of this command. This cannot be null or empty
+	 *
+	 */
+	public void register(String namespace) {
+		if (CommandAPIBukkit.get().isInvalidNamespace(this.meta.commandName, namespace)) {
+			super.register();
+			return;
+		}
```
willkroboth, line 37: We should make sure that no one sets the default namespace to an empty String on Bukkit. Otherwise, this plugin triggers a `StackOverflow` here.

```java
@Override
public void onLoad() {
    CommandAPI.onLoad(new CommandAPIBukkitConfig(this).setNamespace(""));
    new CommandAPICommand("test").register();
}
```

Exception: https://paste.gg/p/anonymous/b3070f86af3b4b57a5055eb86e289a8a

The same thing can happen using a `CommandTree`.

It would probably also be a good idea to add tests for this code path to make sure an invalid namespace really is replaced with the default namespace, and that the default namespace can be properly configured using `setNamespace` and `usePluginNamespace`.

willkroboth (changes requested):

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -742,6 +748,23 @@ public void writeDispatcherToFile() {
 		}
 	}
 
+	<CommandSource> LiteralCommandNode<CommandSource> namespaceNode(LiteralCommandNode<CommandSource> original, String namespace) {
```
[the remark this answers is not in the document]
DerEchtePilz: Dealing with the rest of this review later, I just didn't want to use <Source> here because that would overwrite the class' parameter.
It is necessary to define an explicit parameter here because I think otherwise that resulted in an error. I did write this a few days ago so I am not sure about that but it definitely was a problem. I personally wouldn't change the parameter name here.

Making it static does make sense but I don't think calling `CommandAPIHandler.getInstance()` is a problem.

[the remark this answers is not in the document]
willkroboth: Ah yeah, that's fair. 

https://github.com/JorelAli/CommandAPI/blob/d9ec96fe9c71271aae448cdaf2ba2c9a04829ebe/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java#L100-L116

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -343,15 +346,17 @@ Object parseArgument(CommandContext<Source> cmdCtx, String key, Argument value,
...
+		String namespacedCommand = namespace.isEmpty()
+			? commandName.toLowerCase()
+			: namespace.toLowerCase() + ":" + commandName.toLowerCase();
+		if (registeredPermissions.containsKey(namespacedCommand)) {
+			permission = registeredPermissions.get(namespacedCommand);
 		} else {
-			registeredPermissions.put(commandName.toLowerCase(), permission);
+			registeredPermissions.put(namespacedCommand, permission);
+			// The first command to be registered determines the permission for the `commandName` version of the command
+			registeredPermissions.putIfAbsent(commandName, permission);
```
JorelAli, line 359: `commandName` is case-sensitive in this map - should this have a `.toLowerCase()` here?

JorelAli (approved):

willkroboth (approved):

PR #474, opened 2023-07-24
Rename Adventure arguments

DerEchtePilz:
This essentially makes these changes to make arguments that deal with the same Minecraft argument type have identical names:

- Spigot-related argument changes:

```patch
- dev.jorel.commandapi.arguments.ChatArgument
- dev.jorel.commandapi.arguments.ChatColorArgument
- dev.jorel.commandapi.arguments.ChatComponentArgument

+ dev.jorel.commandapi.arguments.spigot.ChatArgument
+ dev.jorel.commandapi.arguments.spigot.ChatColorArgument
+ dev.jorel.commandapi.arguments.spigot.ChatComponentArgument
```

- Adventure-related argument changes:

```patch
- dev.jorel.commandapi.arguments.AdventureChatArgument
- dev.jorel.commandapi.arguments.AdventureChatColorArgument
- dev.jorel.commandapi.arguments.AdventureChatComponentArgument

+ dev.jorel.commandapi.arguments.adventure.ChatArgument
+ dev.jorel.commandapi.arguments.adventure.ChatColorArgument
+ dev.jorel.commandapi.arguments.adventure.ChatComponentArgument
```

TODO:
- [x] Replace instances of AdventureXArgument with XArgument
- [x] Update Kotlin DSL

Issue #406, opened 2023-01-17
The CommandAPI should use MinecraftServer#getCommands instead of MinecraftServer.vanillaCommandDispatcher


JorelAli:
### Description

As per [this comment](https://github.com/PaperMC/Paper/pull/8235#issuecomment-1369131980):
> ## People using nms
> **Please move away from `MinecraftServer.vanillaCommandDispatcher`, instead, use `MinecraftServer#getCommands()`.**
> 
> This change supports the internal dispatcher being reloaded. We will need to figure out some api in order to properly support re-registering commands under some kind of lifecycle api.

### Expected code

Replace the use of `MINECRAFT_SERVER.vanillaCommandDispatcher` in NMS implementations of `NMS#getBrigadierDispatcher`
https://github.com/JorelAli/CommandAPI/blob/2efd3a3ffdee797ab8b2a4b12784c9e5741d197a/commandapi-nms/commandapi-1.19.3/src/main/java/dev/jorel/commandapi/nms/NMS_1_19_3_R2.java#L395-L398

with `MINECRAFT_SERVER.getCommands()`:

```java
@Override
public final com.mojang.brigadier.CommandDispatcher<CommandSourceStack> getBrigadierDispatcher() {
    return MINECRAFT_SERVER.getCommands().getDispatcher();
}
```

willkroboth:
I notice that comment is from a Paper PR. If this change was applied, would the CommandAPI still work on Spigot? If not, I suppose we would just need to migrate calls to `getBrigadierDispatcher` over to [`PaperImplementations`](https://github.com/JorelAli/CommandAPI/blob/dev/dev/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperImplementations.java).

Also, does this change apply to Paper right now, or only after that PR is merged?

JorelAli:
> I notice that comment is from a Paper PR. If this change was applied, would the CommandAPI still work on Spigot? If not, I suppose we would just need to migrate calls to `getBrigadierDispatcher` over to [`PaperImplementations`](https://github.com/JorelAli/CommandAPI/blob/dev/dev/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperImplementations.java).
> 
> Also, does this change apply to Paper right now, or only after that PR is merged?

[...]

PR #517, opened 2024-01-19
Paper CommandAPI

DerEchtePilz:
This PR refactors parts of the CommandAPI so the Adventure library and Bungee library aren't used in the same .jar's anymore and  Paper API and Spigot API also aren't used together anymore.
In order to do that, `commandapi-bukkit-core` only serves as a tool to access NMS so it can be shared and doesn't have to be implemented again for Paper and Spigot.
The new usable platforms are `commandapi-paper-(core/shade/plugin)` and `commandapi-spigot-(core/shade/plugin)`.

This still requires a lot of work. I am fairly confident that these changes should already make a fairly functional CommandAPI for Paper and Spigot, however, there still is a lot of work left.

This should also invalidate #474 as this refactor puts them into different modules

TODO's:
- [ ] Potentially come up with API to make use of Paper's Brigadier API. That would include:
  - No namespace handling for commands like these
  - Automatic plugin association
  - Probably would require changes to how commands are built or separate API that would make this easier
- [x] Update build script
- [x] Remove `PaperImplementations.java`
- [x] Work through `CommandAPIBukkit` and implement in `CommandAPIPaper` and `CommandAPISpigot` respectively.
- [x] Add Kotlin modules for Spigot and Paper (Component related arguments)
- [x] Make `commandapi-annotations` platform specific (Component related arguments)
- [x] Update documentation code to include at least *a* platform that provides component related arguments
- [x] ~~Update tests to use `CommandAPIPaper` or `CommandAPISpigot`. Maybe even both, share tests and implement component related arguments for both platforms~~ Nope, cancelling those. The original tests are still present but they won't work. After trying to port tests two times I do not really know what to do so if someone else at some point, possibly after this is merged, has an idea this could and should be implemented.
- [x] Update Velocity, apparently
- [x] Deal with #406 for the Paper-specific version
- [x] Make sure the `minecraft` namespace exists for commands on Paper
- [x] Documentation and changelog
- [x] Test Minecraft versions with the plugin only, shade version and non-shade version:
  - [x] 1.16.5
  - [x] 1.17
  - [x] 1.17.1
  - [x] 1.18, 1.18.1
  - [x] 1.18.2
  - [x] 1.19 
  - [x] 1.19.1, 1.19.2
  - [x] 1.19.3
  - [x] 1.19.4
  - [x] 1.20, 1.20.1
  - [x] 1.20.2
  - [x] 1.20.3/4
  - [x] 1.20.5/6 (On Paper, use the CommandAPI using `commandapi-paper-shade-mojang-mapped`)
  - [x] 1.21 (On Paper, use the CommandAPI using `commandapi-paper-shade-mojang-mapped`)


willkroboth (changes requested):
Just an initial look at some things.

The hierarchy of Bukkit and Spigot/Paper NMS feels weird to me, though I can't quite say what to do about it. For example, [these methods](https://github.com/JorelAli/CommandAPI/blob/283788da7b920e2a07d99510a7c45a499988bfb1/commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/CommandAPIPaper.java#L260-L293) in CommandAPIPaper need to call the static `getBukkit` method to get the generics to match up. Implementations of NMS call up `CommandAPIBukkit#wrapCommandSender`, which [passes it over](https://github.com/JorelAli/CommandAPI/blob/283788da7b920e2a07d99510a7c45a499988bfb1/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L290) to `BukkitPlatform#wrapCommandSender`. Nothing major, but I just wonder if there is a cleaner way. Maybe I'll do some poking around on my own and see if I find anything.

DerEchtePilz:
About your comment, yeah definitely weird at some places.
The main goal was to share as much code as possible while changing inheritances around.
For example, the methods you linked to in the first link are all defined in `CommandAPIPlatform` but as `CommandAPIBukkit` is not a `CommandAPIPlatform` anymore, I left the methods there so `CommandAPIPaper` and `CommandAPISpigot` can call these methods.
The `wrapCommandSender` method is another case of this weird "needs to be implemented platform specific but also needs to be accessible from `CommandAPIBukkit`" state since the Bukkit NMS classes call this method.
Definitely let me know if you come up with a way to handle this better/different, but this definitely accomplishes the goal of making the Bukkit modules an interface for the core NMS features and implementations.

commandapi-annotations/pom.xml
```diff
@@ -54,18 +55,6 @@
...
-						<goals>
-							<goal>testCompile</goal>
-						</goals>
-						<configuration>
-							<annotationProcessors>
-								<annotationProcessor>dev.jorel.commandapi.annotations.Annotations</annotationProcessor>
-							</annotationProcessors>
-						</configuration>
-					</execution>
-				</executions>
```
willkroboth, line 68: If this build plugin no longer has any `executions`, does it need to be declared anymore? Could this removal be expanded to also remove the whole `build` section?

commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/CommandAPIPaper.java
```diff
@@ -0,0 +1,244 @@
...
+			return new BukkitProxiedCommandSender(proxy);
+		}
+		if (sender instanceof RemoteConsoleCommandSender remote) {
+			return new BukkitRemoteConsoleCommandSender(remote);
+		}
+		final Class<? extends CommandSender> FeedbackForwardingSender = feedbackForwardingCommandSender;
+		if (FeedbackForwardingSender.isInstance(sender)) {
+			// We literally cannot type this at compile-time, so let's use a placeholder CommandSender instance
+			return new BukkitFeedbackForwardingCommandSender<CommandSender>(FeedbackForwardingSender.cast(sender));
+		}
```
willkroboth, line 213: Storing the class in a local variable (same for `NullCommandSender`) probably isn't necessary when the field is in the same class it is being used (previously the field came from `PaperImplementations` but was used in `CommandAPIBukkit`).

```suggestion
		if (this.feedbackForwardingCommandSender.isInstance(sender)) {
			// We literally cannot type this at compile-time, so let's use a placeholder CommandSender instance
			return new BukkitFeedbackForwardingCommandSender<CommandSender>(this.feedbackForwardingCommandSender.cast(sender));
		}
```

commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/arguments/CustomArgument.java
```diff
@@ -0,0 +1,377 @@
...
+
+	/**
+	 * An exception used to create command-related errors for the CustomArgument
+	 */
+	@SuppressWarnings("serial")
+	public static class CustomArgumentException extends Exception {
+
+		private Component errorComponent = null;
+		private String errorMessage = null;
+		private MessageBuilder errorMessageBuilder = null;
```
willkroboth, line 223: Perhaps instead of copying the whole `CustomArgument` class, which is otherwise identical Spigot/Paper, the inner `CustomArgumentException` class could be extracted and given the platform-specific component-type error messages.

commandapi-platforms/commandapi-paper/commandapi-paper-nms/commandapi-paper-nms-common/src/main/java/dev/jorel/commandapi/nms/PaperNMS_Common.java
```diff
@@ -0,0 +1,30 @@
...
+import net.minecraft.commands.arguments.MessageArgument;
+
+import static dev.jorel.commandapi.preprocessor.Unimplemented.REASON.VERSION_SPECIFIC_IMPLEMENTATION;
+
+public abstract class PaperNMS_Common extends CommandAPIPaper<CommandSourceStack> {
+
+	@Override
+	public final Component getChat(CommandContext<CommandSourceStack> cmdCtx, String key) throws CommandSyntaxException {
+		return GsonComponentSerializer.gson().deserialize(net.minecraft.network.chat.Component.Serializer.toJson(MessageArgument.getMessage(cmdCtx, key)));
+	}
```
willkroboth, line 20: This method should be overriden in 1.20.5+ because `Serializer#toJson` was changed in that version to require the `COMMAND_BUILD_CONTEXT`. Currently, with this code:

```java
@Override
public void onEnable() {
    // Plugin startup logic
    CommandAPI.onLoad(new CommandAPIPaperConfig(this).verboseOutput(true));

    CommandAPI.onEnable();

    new CommandAPICommand("test")
            .withArguments(new ChatArgument("chat"))
            .executesPlayer(info -> {
                Component component = info.args().getUnchecked("chat");

                info.sender().sendMessage(component);
            })
            .register();
}
```

Running `/test Hello` in the console gives the following error, causing the command to fail.

```
Unexpected exception while parsing console command "test Hello"
org.bukkit.command.CommandException: Unhandled exception executing 'test Hello' in org.bukkit.craftbukkit.command.VanillaCommandWrapper(test)
        at org.bukkit.command.SimpleCommandMap.dispatch(SimpleCommandMap.java:175) ~[paper-mojangapi-1.21.1-R0.1-SNAPSHOT.jar:?]
        at org.bukkit.craftbukkit.CraftServer.dispatchCommand(CraftServer.java:1000) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at org.bukkit.craftbukkit.CraftServer.dispatchServerCommand(CraftServer.java:985) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.server.dedicated.DedicatedServer.handleConsoleInputs(DedicatedServer.java:526) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.server.dedicated.DedicatedServer.tickChildren(DedicatedServer.java:474) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.server.MinecraftServer.tickServer(MinecraftServer.java:1598) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1304) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:330) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
Caused by: java.lang.NoSuchMethodError: 'java.lang.String net.minecraft.network.chat.Component$Serializer.toJson(net.minecraft.network.chat.Component)'
        at BukkitTestPlugin-1.0-SNAPSHOT.jar/dev.jorel.commandapi.nms.PaperNMS_Common.getChat(PaperNMS_Common.java:19) ~[BukkitTestPlugin-1.0-SNAPSHOT.jar:?]
        at BukkitTestPlugin-1.0-SNAPSHOT.jar/dev.jorel.commandapi.arguments.ChatArgument.parseArgument(ChatArgument.java:80) ~[BukkitTestPlugin-1.0-SNAPSHOT.jar:?]
        at BukkitTestPlugin-1.0-SNAPSHOT.jar/dev.jorel.commandapi.arguments.ChatArgument.parseArgument(ChatArgument.java:46) ~[BukkitTestPlugin-1.0-SNAPSHOT.jar:?]
        at BukkitTestPlugin-1.0-SNAPSHOT.jar/dev.jorel.commandapi.CommandAPIHandler.parseArgument(CommandAPIHandler.java:319) ~[BukkitTestPlugin-1.0-SNAPSHOT.jar:?]
        at BukkitTestPlugin-1.0-SNAPSHOT.jar/dev.jorel.commandapi.CommandAPIHandler.argsToCommandArgs(CommandAPIHandler.java:291) ~[BukkitTestPlugin-1.0-SNAPSHOT.jar:?]
        at BukkitTestPlugin-1.0-SNAPSHOT.jar/dev.jorel.commandapi.CommandAPIHandler.lambda$generateCommand$0(CommandAPIHandler.java:195) ~[BukkitTestPlugin-1.0-SNAPSHOT.jar:?]
        at com.mojang.brigadier.context.ContextChain.runExecutable(ContextChain.java:73) ~[brigadier-1.3.10.jar:?]
        at net.minecraft.commands.execution.tasks.ExecuteCommand.execute(ExecuteCommand.java:31) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.commands.execution.tasks.ExecuteCommand.execute(ExecuteCommand.java:19) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.commands.execution.UnboundEntryAction.lambda$bind$0(UnboundEntryAction.java:8) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.commands.execution.CommandQueueEntry.execute(CommandQueueEntry.java:5) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.commands.execution.ExecutionContext.runCommandQueue(ExecutionContext.java:103) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.commands.Commands.executeCommandInContext(Commands.java:443) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.commands.Commands.performCommand(Commands.java:350) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.commands.Commands.performCommand(Commands.java:337) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at net.minecraft.commands.Commands.performPrefixedCommand(Commands.java:327) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at org.bukkit.craftbukkit.command.VanillaCommandWrapper.execute(VanillaCommandWrapper.java:55) ~[paper-1.21.1.jar:1.21.1-41-fcedb49]
        at org.bukkit.command.SimpleCommandMap.dispatch(SimpleCommandMap.java:165) ~[paper-mojangapi-1.21.1-R0.1-SNAPSHOT.jar:?]
        ... 8 more
```

Add the `@Overriden` annotation to this method matches the current implementation on `dev/dev` in [`NMS_Common`](https://github.com/JorelAli/CommandAPI/blob/c3ff409407ba18f3b0561051bdc99b1f89ca1678/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-nms-common/src/main/java/dev/jorel/commandapi/nms/NMS_Common.java#L310-L314). `dev/paper/restructure-experiment` currently [does this](https://github.com/JorelAli/CommandAPI/blob/d9c1f0d6a85cd94ab7d2d20682c12ad20684ea54/commandapi-platforms/commandapi-paper/commandapi-paper-nms/commandapi-paper-nms-common/src/main/java/dev/jorel/commandapi/nms/PaperNMS_Common.java#L18-L35), and I've verified that the Paper `ChatArgument` over there works on real servers.

Addtionally, `getChatComponent` can be implemented here, with the same exception of needed to be overridden in 1.20.5+.

commandapi-platforms/commandapi-paper/commandapi-paper-nms/commandapi-paper-nms-common/src/main/java/dev/jorel/commandapi/nms/PaperNMS_CommonWithFunctions.java
```diff
@@ -0,0 +1,5 @@
+package dev.jorel.commandapi.nms;
+
+public abstract class PaperNMS_CommonWithFunctions extends PaperNMS_Common {
+
+}
```
willkroboth, line 5: This class doesn't seem necessary. Instead of extending `PaperNMS_CommonWithFunction`, subclasses could just extend `PaperNMS_Common` directly. `dev/paper/restructure-experiment` has [done this](https://github.com/JorelAli/CommandAPI/commit/666a422416948bc756ed6331768563da74a83e02#diff-7672b0644b015eb0e0888ed2568021f5e3c2fd9b369cff243753ed72707a36d4L1-L5) without any problems.

commandapi-platforms/commandapi-spigot/commandapi-spigot-core/src/main/java/dev/jorel/commandapi/nms/SpigotNMS.java
```diff
@@ -0,0 +1,20 @@
...
+
+public interface SpigotNMS<CommandListenerWrapper> {
+
+	BaseComponent[] getChat(CommandContext<CommandListenerWrapper> cmdCtx, String key) throws CommandSyntaxException;
+
+	ChatColor getChatColor(CommandContext<CommandListenerWrapper> cmdCtx, String key);
+
+	BaseComponent[] getChatComponent(CommandContext<CommandListenerWrapper> cmdCtx, String key);
+
+	NMS<?> bukkitNMS();
```
willkroboth, line 18: This (and the same method in PaperNMS) could use the generic parameter to avoid a [cast in `CommandAPISpigot`](https://github.com/JorelAli/CommandAPI/blob/7d2a622ab856eabf2393d8c35bee309bad646177/commandapi-platforms/commandapi-spigot/commandapi-spigot-core/src/main/java/dev/jorel/commandapi/CommandAPISpigot.java#L36).

```suggestion
	NMS<CommandListenerWrapper> bukkitNMS();
```

commandapi-platforms/commandapi-paper/commandapi-paper-nms/commandapi-paper-nms-common/src/main/java/dev/jorel/commandapi/nms/PaperNMS_Common.java
```diff
@@ -0,0 +1,30 @@
...
+
+	@Override
+	@Unimplemented(because = VERSION_SPECIFIC_IMPLEMENTATION, from = "ofExact", to = "namedColor", in = "NamedTextColor", introducedIn = "Adventure 4.10.0", info = "1.18 uses Adventure 4.9.3. 1.18.2 uses Adventure 4.11.0")
+	public abstract NamedTextColor getChatColor(CommandContext<CommandSourceStack> cmdCtx, String key);
+
+	@Override
+	public NMS<?> bukkitNMS() {
+		return null;
+	}
```
willkroboth, line 29: This seems odd. This probably shouldn't return `null`. This method declaration should probably be removed so it can just be inherited normally in the subclasses.

```suggestion
```

commandapi-platforms/commandapi-spigot/commandapi-spigot-core/src/main/java/dev/jorel/commandapi/CommandAPISpigot.java
```diff
@@ -0,0 +1,137 @@
...
+	@SuppressWarnings("unchecked")
+	protected CommandAPISpigot() {
+		this.nms = (NMS<Source>) bukkitNMS();
+		CommandAPISpigot.spigot = this;
+		setInstance(this);
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <Source> CommandAPISpigot<Source> getSpigot() {
+		return (CommandAPISpigot<Source>) spigot;
```
willkroboth, line 43: Should this have a `hey, you're calling this too early` check like there is in [`CommandAPIBukkit`](https://github.com/JorelAli/CommandAPI/blob/7d2a622ab856eabf2393d8c35bee309bad646177/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java#L68)?

commandapi-platforms/commandapi-spigot/commandapi-spigot-core/src/main/java/dev/jorel/commandapi/CommandAPISpigot.java
```diff
@@ -0,0 +1,137 @@
...
+
+	@Override
+	public <T extends CommandAPIBukkitConfig<T>> void onLoad(CommandAPIBukkitConfig<T> config) {
+		if (config instanceof CommandAPISpigotConfig spigotConfig) {
+			CommandAPISpigot.setInternalConfig(new InternalSpigotConfig(spigotConfig));
+		} else {
+			CommandAPI.logError("CommandAPIBukkit was loaded with non-Bukkit config!");
+			CommandAPI.logError("Attempts to access Bukkit-specific config variables will fail!");
+		}
+		onLoad();
```
willkroboth, line 62: It would be clearer this calls the method in `CommandAPIBukkit` if it used the `super` keyword. At the moment, I was like, "Wait, where's the no-parameter `onLoad` method in `CommandAPISpigot`?" when I saw this.

```suggestion
		super.onLoad();
```

commandapi-platforms/commandapi-spigot/commandapi-spigot-nms/commandapi-spigot-nms-common/src/main/java/dev/jorel/commandapi/nms/SpigotNMS_Common.java
```diff
@@ -0,0 +1,28 @@
...
+import dev.jorel.commandapi.CommandAPISpigot;
+import net.md_5.bungee.api.chat.BaseComponent;
+import net.minecraft.commands.CommandSourceStack;
+import net.minecraft.commands.arguments.ColorArgument;
+import org.bukkit.ChatColor;
+
+public abstract class SpigotNMS_Common extends CommandAPISpigot<CommandSourceStack> {
+
+	@Override
+	public abstract BaseComponent[] getChat(CommandContext<CommandSourceStack> cmdCtx, String key) throws CommandSyntaxException;
```
willkroboth, line 14: `getChat` (and `getChatComponent`) could be implemented here, as long as they are `@Overriden` in 1.20.5+, since `Serializer#toJson` needs the `COMMAND_BUILD_CONTEXT` then. 

[`NMS_Common` on `dev/dev`](https://github.com/JorelAli/CommandAPI/blob/c3ff409407ba18f3b0561051bdc99b1f89ca1678/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-nms-common/src/main/java/dev/jorel/commandapi/nms/NMS_Common.java#L348-L352) and `dev/paper/restructure-experiment` currently [does this](https://github.com/JorelAli/CommandAPI/blob/d9c1f0d6a85cd94ab7d2d20682c12ad20684ea54/commandapi-platforms/commandapi-spigot/commandapi-spigot-nms/commandapi-spigot-nms-common/src/main/java/dev/jorel/commandapi/nms/SpigotNMS_Common.java#L16-L33).

willkroboth (changes requested):

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
```diff
@@ -429,11 +287,6 @@ public void registerPermission(String string) {
 		}
 	}
 
-	@Override
-	@Unimplemented(because = REQUIRES_MINECRAFT_SERVER)
-	public abstract SuggestionProvider<Source> getSuggestionProvider(SuggestionProviders suggestionProvider);
-
-	@Override
```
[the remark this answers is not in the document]
DerEchtePilz: Probably a remnant from when `CommandAPIBukkit` wasn't a `CommandAPIPlatform`

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/nms/NMS.java
```diff
@@ -70,61 +76,63 @@
...
-import net.kyori.adventure.text.format.NamedTextColor;
-import net.md_5.bungee.api.chat.BaseComponent;
 
-public interface NMS<CommandListenerWrapper> {
+import static dev.jorel.commandapi.preprocessor.Unimplemented.REASON.REQUIRES_CRAFTBUKKIT;
+import static dev.jorel.commandapi.preprocessor.Unimplemented.REASON.REQUIRES_CSS;
+import static dev.jorel.commandapi.preprocessor.Unimplemented.REASON.REQUIRES_MINECRAFT_SERVER;
+import static dev.jorel.commandapi.preprocessor.Unimplemented.REASON.VERSION_SPECIFIC_IMPLEMENTATION;
+
+public abstract class NMS<CommandListenerWrapper> {
```
willkroboth, line 85: Just a note that NMS could still be an interface. Doesn't really matter, but could reduce the diff here so the important changes are easier to see.

commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/CommandAPIPaper.java
```diff
@@ -0,0 +1,308 @@
...
+		public void severe(String message) {
+			componentLogger.error(message);
+		}
+
+		@Override
+		public void severe(String message, Throwable exception) {
+			componentLogger.error(message, exception);
+		}
+	}
+
```
willkroboth, line 307: There is a utility method for turning a slf4j logger (which `ComponentLogger` extends) into a `CommandAPILogger`.

```suggestion
	@SuppressWarnings("ConstantValue")
	@Override
	public CommandAPILogger getLogger() {
		if (Bukkit.getServer() != null) {
			return super.getLogger();
		}
		if (bootstrapLogger == null) {
			bootstrapLogger = CommandAPILogger.fromSlf4jLogger(ComponentLogger.logger("CommandAPI"));
		}
		return bootstrapLogger;
	}
```

commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -0,0 +1,155 @@
...
+		CommandAPIHandler.getInstance().writeDispatcherToFile();
+	}
+
+	@Override
+	public void preReloadDataPacks() {
+		reloadHelpTopics.run();
+		CommandAPIBukkit.get().updateHelpForCommands(CommandAPI.getRegisteredCommands());
+	}
+
+	void registerLifecycleEvent(boolean bootstrap) {
```
willkroboth, line 111: When this method is called, the argument is always `Bukkit.getServer() == null`. It seems like this could be moved into this method.

```suggestion
	void registerLifecycleEvent() {
		boolean bootstrap = Bukkit.getServer() == null;
```

commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -0,0 +1,155 @@
...
+		if (bootstrap && !lifecycleEventRegistered[0]) {
+			BootstrapContext context = (BootstrapContext) CommandAPIPaper.getPaper().getLifecycleEventOwner();
+			lifecycleEventRegistered[0] = true;
```
willkroboth, line 114: Is the `lifecycleEventRegistered` variable needed? It seems like this method is only called once before/after bootstrap.

commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -0,0 +1,155 @@
...
+		}
+		if (!bootstrap && !lifecycleEventRegistered[1]) {
+			JavaPlugin plugin = (JavaPlugin) CommandAPIPaper.getPaper().getLifecycleEventOwner();
+			lifecycleEventRegistered[1] = true;
+			plugin.getLifecycleManager().registerEventHandler(LifecycleEvents.COMMANDS, event -> {
+				for (CommandNode<CommandSourceStack> commandNode : pluginDispatcher.getRoot().getChildren()) {
+					LiteralCommandNode<CommandSourceStack> node = (LiteralCommandNode<CommandSourceStack>) commandNode;
+					event.registrar().register(node, getDescription(node.getLiteral()));
+				}
+			});
```
willkroboth, line 131: This method may be easier to understand if this common code block was extracted.

```java
void registerLifecycleEvent() {
	boolean bootstrap = Bukkit.getServer() == null;

	if (bootstrap) {
		BootstrapContext context = (BootstrapContext) CommandAPIPaper.getPaper().getLifecycleEventOwner();
		registerLifecycleEvent(context.getLifecycleManager(), bootstrapDispatcher);
	} else {
		JavaPlugin plugin = (JavaPlugin) CommandAPIPaper.getPaper().getLifecycleEventOwner();
		registerLifecycleEvent(plugin.getLifecycleManager(), pluginDispatcher);
	}
}

private void registerLifecycleEvent(LifecycleEventManager<?> lifecycleManager, CommandDispatcher<CommandSourceStack> dispatcher) {
	lifecycleManager.registerEventHandler(LifecycleEvents.COMMANDS, event -> {
		for (CommandNode<CommandSourceStack> commandNode : dispatcher.getRoot().getChildren()) {
			LiteralCommandNode<CommandSourceStack> node = (LiteralCommandNode<CommandSourceStack>) commandNode;
			event.registrar().register(node, getDescription(node.getLiteral()));
		}
	});
}
```

It took me a while to notice the one difference between the lambda methods :P

commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/arguments/AsyncPlayerProfileArgument.java
```diff
@@ -1,34 +1,37 @@
...
 import dev.jorel.commandapi.CommandAPIBukkit;
+import dev.jorel.commandapi.CommandAPIPaper;
 import dev.jorel.commandapi.executors.CommandArguments;
 import org.bukkit.OfflinePlayer;
 
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
 /**
  * An argument that represents the Bukkit OfflinePlayer object with asynchronous support.
```
willkroboth, line 15: The Javadocs here reference the old objects still. Same in Spigot.

commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/arguments/AsyncPlayerProfileArgument.java
```diff
@@ -1,34 +1,37 @@
...
  *
  * @since 9.7.1
  */
-public class AsyncOfflinePlayerArgument extends SafeOverrideableArgument<CompletableFuture<OfflinePlayer>, OfflinePlayer> {
+public class AsyncPlayerProfileArgument extends SafeOverrideableArgument<CompletableFuture<List<PlayerProfile>>, PlayerProfile> {
```
willkroboth, line 19:
```suggestion
public class AsyncPlayerProfileArgument extends SafeOverrideableArgument<CompletableFuture<PlayerProfile>, PlayerProfile> {
```

Should we return a list of `PlayerProfiles`? If this always resolves to a list with a single element, we should probably just return that element. Or does this actually give multiple players?

commandapi-platforms/commandapi-paper/commandapi-paper-plugin/src/main/java/dev/jorel/commandapi/CommandAPIMain.java
```diff
@@ -0,0 +1,157 @@
...
+				return javaPlugin;
+			} else {
+				new InvalidPluginException("Plugin " + pluginName + " is not a JavaPlugin!").printStackTrace();
+			}
+		} else {
+			new InvalidPluginException("Could not find a plugin " + pluginName + "! Has it been loaded properly?")
+				.printStackTrace();
+		}
+		return null;
+	}
```
willkroboth, line 133: Can this logic be shared with Spigot? It seems identical.

commandapi-platforms/commandapi-paper/commandapi-paper-plugin/src/main/resources/config.yml
```diff
@@ -0,0 +1,83 @@
+################################################################################
```
willkroboth, line 1: Is this file necessary anymore? The config should be automatically generated now.

commandapi-platforms/commandapi-spigot/commandapi-spigot-core/src/main/java/dev/jorel/commandapi/CommandAPISpigot.java
```diff
@@ -0,0 +1,136 @@
...
+		Bukkit.getScheduler().scheduleSyncDelayedTask(CommandAPIBukkit.get().plugin, () -> {
+			// Sort out permissions after the server has finished registering them all
+			CommandAPIBukkit.get().getCommandRegistrationStrategy().runTasksAfterServerStart();
+			if (!getConfiguration().skipReloadDatapacks()) {
+				CommandAPIBukkit.get().reloadDataPacks();
+			}
+			CommandAPIBukkit.get().updateHelpForCommands(CommandAPI.getRegisteredCommands());
+		}, 0L);
+
+		CommandAPIBukkit.get().stopCommandRegistrations();
```
willkroboth, line 84: `CommandAPISpigot` extends `CommandAPIBukkit`, so I don't think `CommandAPIBukkit.get()` is needed anymore. Same in Paper.

commandapi-platforms/commandapi-spigot/commandapi-spigot-core/src/main/java/dev/jorel/commandapi/arguments/AsyncPlayerProfileArgument.java
```diff
@@ -0,0 +1,51 @@
...
+import dev.jorel.commandapi.executors.CommandArguments;
+import org.bukkit.profile.PlayerProfile;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * An argument that represents the Bukkit OfflinePlayer object with asynchronous support.
+ *
+ * @since 9.7.1
```
willkroboth, line 16: I think we should update these since tags? These `PlayerProfile` classes are different from the ones they replaced.

commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/arguments/AsyncPlayerProfileArgument.java
DerEchtePilz: It can return multiple profiles when using selectors.

willkroboth: Ah, so I suppose these javadocs should be updated instead

https://github.com/CommandAPI/CommandAPI/blob/63d55a5d587773ff947fdae56a0a6ef88b62e5f8/commandapi-platforms/commandapi-paper/commandapi-paper-core/src/main/java/dev/jorel/commandapi/arguments/AsyncPlayerProfileArgument.java#L21-L26

PR #521, opened 2024-02-23
Velocity Documentation

DerEchtePilz:
This is supposed to add a bit of Velocity documentation to the documentation.

Feel free to suggest anything else that you think should be added!

### DIFF
--- a/commandapi-documentation-velocity-code/pom.xml
+++ b/commandapi-documentation-velocity-code/pom.xml
@@ -3,14 +3,13 @@
 		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
-
 	<parent>
-		<artifactId>commandapi</artifactId>
 		<groupId>dev.jorel</groupId>
+		<artifactId>commandapi</artifactId>
 		<version>9.4.0-SNAPSHOT</version>
 	</parent>
 
-	<artifactId>commandapi-documentation-code</artifactId>
+	<artifactId>commandapi-documentation-velocity-code</artifactId>
 
 	<repositories>
 		<repository>
@@ -43,31 +42,14 @@
 			<scope>provided</scope>
 		</dependency>
 
-		<!-- Bukkit dependencies -->
 		<dependency>
-			<groupId>net.kyori</groupId>
-			<artifactId>adventure-platform-bukkit</artifactId>
-			<version>4.2.0</version>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>io.papermc.paper</groupId>
-			<artifactId>paper-api</artifactId>
-			<version>${paper.version}</version>
+			<groupId>com.velocitypowered</groupId>
+			<artifactId>velocity-api</artifactId>
+			<version>3.3.0-SNAPSHOT</version>
 			<scope>provided</scope>
 		</dependency>
 
 		<!-- CommandAPI -->
-		<dependency>
-			<groupId>dev.jorel</groupId>
-			<artifactId>commandapi-bukkit-core</artifactId>
-			<version>${project.version}</version>
-		</dependency>
-		<dependency>
-			<groupId>dev.jorel</groupId>
-			<artifactId>commandapi-bukkit-kotlin</artifactId>
-			<version>${project.version}</version>
-		</dependency>
 		<dependency>
 			<groupId>dev.jorel</groupId>
 			<artifactId>commandapi-velocity-core</artifactId>
@@ -113,4 +95,5 @@
 			</plugin>
 		</plugins>
 	</build>
+
 </project>
--- a/commandapi-documentation-velocity-code/pom.xml
+++ b/commandapi-documentation-velocity-code/pom.xml
@@ -20,11 +20,6 @@
 			<id>papermc</id>
 			<url>https://papermc.io/repo/repository/maven-public/</url>
 		</repository>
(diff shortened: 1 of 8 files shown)


JorelAli:
I've been thinking of this a bit more - if we're going to have a clear separation between sections and chapters in our documentation, would it be easier to maintain (and more well-structured) to format the directory structure as well? I believe mdBook allows directories for its `SUMMARY.md` file, so we could have something like...

```
.
├── velocity
│   └── velocity_intro.md
└── spigot
    └── (everything else)
```

Just a thought. Not sure how relevant this is at this stage, but if we continue branching out to different platforms (fabric, forge, sponge, etc.), it's probably worth considering.

willkroboth:
Maybe we could also bring the `documentation-code` structure together with the `docssrc` files. Something like this maybe?

```
commandapi-documentation/
├─ pom.xml
├─ docssrc/
│  └─ intro.md
├─ bukkit/
│  ├─ docssrc/
│  │  ├─ bukkit_intro.md
│  │  ├─ arguments/
│  │  │  ├─ argument_whatever1.md
│  │  │  ├─ argument_whatever2.md
│  │  │  └─ ...
│  │  └─ annotations/
│  │     └─ ...
│  └─ commandapi-documentation-bukkit-code/
│     ├─ src/main/.../
│     └─ pom.xml
└─ velocity/
   ├─ docssrc/
   │  └─ velocity_intro.md
   └─ commandapi-documenatation-velocity-code/
      ├─ src/main/.../
      └─ pom.xml
```

That way, for example, when you're working with the bukkit documentation pages `bukkit/docssrc`, you're simply one directory away from documentation examples `bukkit/commandapi-documentation-bukkit-code`.

Or maybe we could completely merge the `documentation-code` and `docssrc` files like so:

```
commandapi-documenation/
├─ pom.xml
├─ src/
│  └─ main/
│     └─ intro.md
├─ commandapi-documenation-bukkit/
│  ├─ pom.xml
│  └─ src/
│     └─ main/
│        ├─ bukkit_intro.md
│        ├─ arguments/
│        │  ├─ argument1.md
│        │  ├─ argument1.java
│        │  ├─ argument1.kt
│        │  ├─ argument2.md
│        │  ├─ argument2.java
│        │  ├─ argument2.kt
│        │  └─ ...
│        └─ annotations/
│           └─ ...
└─ commandapi-documentation-velocity/
   ├─ pom.xml
   └─ src/
      └─ main/
         ├─ velocity_intro.md
         ├─ randomNumberCommand.java
         ├─ randomNumberCommand.kt
         └─ randomNumberCommandDSL.kt
```
I'm not sure if mdBook, Java, or Kotlin like all those files in the same place, but it would make it very easy to find the files that hold the code that documentation pages reference. Individual code files for each documentation page might be easier to understand than monolithic `Example` classes that hold everything.

But yeah, that's probably not something we need to set up now for a single page about Velocity.

DerEchtePilz:
For this PR, I intentionally paid attention to not change any existing documentation code locations because that would need to be updated in the pages as well which to be honest I deemed to not be within the scope of this PR.
During 9.0.0 development, I sorted the examples alphabetically and with the method they are in you also find the page rather quickly. The class, however, is quite big so splitting that up to a per-page class might also help.
With #517 I am also already trying to introduce platform specific documentation so something like this maybe should wait.

docssrc/src/velocity_intro.md
```diff
@@ -0,0 +1,115 @@
...
+dependencies {
+    compile "dev.jorel:commandapi-velocity-shade:9.4.0-SNAPSHOT"
+}
+```
+
+```gradle,build.gradle.kts
+dependencies {
+    compileOnly("dev.jorel:commandapi-velocity-shade:9.4.0-SNAPSHOT")
+}
+```
```
willkroboth, line 67: Should we add these to the `update.sh` script?

https://github.com/JorelAli/CommandAPI/blob/800372c52ce7b3ca4195224fc1487d31d5cb48d7/update.sh#L11-L22

Maybe not right now. The versions on this documentation page aren't supposed to be exactly up-to-date. They're supposed to be the snapshot version before the newest version.

pom.xml
```diff
@@ -70,10 +70,11 @@
 
 		<!-- Documentation code -->
 		<module>commandapi-documentation-code</module>
+        <module>commandapi-documentation-velocity-code</module>
```
willkroboth, line 73: Not a big deal, but the line here uses spaces while the surrounding lines use tabs. I only noticed because of the change line 76->77, where the only change is going from tabs to spaces.

willkroboth (approved):
Just a few notes. Maybe once we do some more Velocity docs it would make sense to restructure, but this PR definitely achieves the goal of letting people know Velocity CommandAPI exists.

commandapi-documentation-velocity-code/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+			<artifactId>velocity-api</artifactId>
+			<version>3.3.0-SNAPSHOT</version>
+			<scope>provided</scope>
+		</dependency>
+
+		<!-- CommandAPI -->
+		<dependency>
+			<groupId>dev.jorel</groupId>
+			<artifactId>commandapi-velocity-core</artifactId>
+			<version>9.4.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
DerEchtePilz: Can this be changed to  `${project.version}`? I wanted to do that initially, but then thought about the update stuff and wondered if that was safe to do.

DerEchtePilz:
> Maybe once we do some more Velocity docs it would make sense to restructure

No argument here. Restructuring definitely makes sense but also definitely doesn't fit into this PR.


Also thanks for the few notes. I'll do them all locally but will add you as a co-author for the suggested changes as GitHub would do the same.

Issue #483, opened 2023-08-18
Add a `FlagsArgument`


willkroboth:
### Description

This suggestion comes from a common question on the CommandAPI Discord. Most recently, a [comment from Lear](https://discord.com/channels/745416925924032523/745757702608912556/1142062958801866752) reminded me of this, but there have also been ideas from [TheZequex](https://discord.com/channels/745416925924032523/745757702608912556/1134886180266184737), [Skepter](https://discord.com/channels/745416925924032523/745419648970784821/1106316627928166400), [NextdoorPsycho](https://discord.com/channels/745416925924032523/745757702608912556/1104843561670807644) and probably lots more.

In Vanilla Minecraft, the [`execute` command](https://minecraft.fandom.com/wiki/Commands/execute#Syntax) is very powerful. It has many subcommands that a user can type out in any order, with each subcommand having its own arguments that define how the command's context should be changed. For example, you can type commands like this:

```
execute as @e if block ~ ~ ~ minecraft:air ...
execute if block ~ ~ ~ minecraft:air as @e ...
```

Here, the `as` and `if block` subcommands can go in any order. `as` has an EntitySelectorArgument before it loops back, and `if block` has a LocationArgument followed by a BlockPredicateArgument. You can keep going, adding new subcommands until the `run` subcommand, which allows you to exit the loop and run any other command.

The Brigadier system handles this by allowing redirects to any other node in the CommandDispatch tree. In `/execute`'s case, each subcommand redirects back to the base `execute` node to continue the loop.

The CommandAPI can help a little with writing commands like this, as shown by [this example](https://commandapi.jorel.dev/9.1.0/brigadier.html#example---adding-a-predicate-to-the-execute-command) in the documentation. However, you still need to import `Brigadier` and deal with its classes directly. Additionally, the Brigadier source when using Bukkit is the NMS class `CommandListenerWrapper`, which means you have to deal with raw types when building the commands. I think it's possible to give a bit more help here.

### Expected code

I'll show what I'm thinking with [Lear's usecase](https://discord.com/channels/745416925924032523/745757702608912556/1142062958801866752). Lear wanted to implement filtering for a custom object with syntax like `/findCustomObject sort=random limit=1 distance=..3`, similar to EntitySelector filtering. Using the `FlagsArgument` as I'm calling it, something similar could be possible like so:

```java
new CommandAPICommand("findCustomObject")
        .withArguments(
                new FlagsArgument("filters")
                        .loopingBranch(
[...]

PR #522, opened 2024-02-24
v9.3.0-mod

stumper66:
Added a new option: skipListValidation

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentBuilder.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentBuilder.java
@@ -21,6 +21,7 @@ public class ListArgumentBuilder<T> {
 	private final String nodeName;
 	private final String delimiter;
 	private boolean allowDuplicates = false;
+	private boolean skipListValidation = false;
 
 	/**
 	 * Creates a new ListArgumentBuilder with a specified node name. Defaults the
@@ -53,6 +54,17 @@ public class ListArgumentBuilder<T> {
 	 */
 	public ListArgumentBuilder<T> allowDuplicates(boolean allowDuplicates) {
 		this.allowDuplicates = allowDuplicates;
+		return this;
+	}
+
+	/**
+	 * Whether the list verification should be skipped
+	 *
+	 * @param skipListValidation whether to skip list validation
+	 * @return this list argument builder
+	 */
+	public ListArgumentBuilder<T> skipListValidation(boolean skipListValidation) {
+		this.skipListValidation = skipListValidation;
 		return this;
 	}
 
@@ -161,7 +173,7 @@ public class ListArgumentBuilder<T> {
 			 * @return a {@link ListArgument}
 			 */
 			public ListArgument<T> buildGreedy() {
-				return new ListArgument<>(nodeName, delimiter, allowDuplicates, supplier, mapper);
+				return new ListArgument<>(nodeName, delimiter, allowDuplicates, supplier, mapper, skipListValidation);
 			}
 			
 			/**
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
@@ -134,7 +134,6 @@ public class ListArgumentCommon<T> extends Argument<List> {
 
 		// If the argument's value is in the list of values, include it
 		List<T> list = new ArrayList<>();
-		if (skipListValidation) return list;
 		String argument = cmdCtx.getArgument(key, String.class);
 		String[] strArr = argument.split(Pattern.quote(delimiter));
 		StringReader context = new StringReader(argument);
@@ -157,7 +156,7 @@ public class ListArgumentCommon<T> extends Argument<List> {
 					addedItem = true;
 				}
 			}
-			if(!addedItem) {
+			if(!addedItem && !skipListValidation) {
 				context.setCursor(cursor);
 				throw new SimpleCommandExceptionType(new LiteralMessage("Item is not allowed in list")).createWithContext(context);
 			}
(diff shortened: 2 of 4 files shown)


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
```diff
@@ -132,6 +134,7 @@ public <CommandSourceStack> List<T> parseArgument(CommandContext<CommandSourceSt
 
 		// If the argument's value is in the list of values, include it
 		List<T> list = new ArrayList<>();
+		if (skipListValidation) return list;
```
DerEchtePilz, line 137: I am unsure about this check being here. I mean, despite not checking if only valid values were given, you probably still would want access to the given list in the argument.
I'd suggest using the boolean to disable the exceptions from being thrown instead of using it to return an empty list.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentBuilder.java
```diff
@@ -161,7 +173,7 @@ private ListArgumentBuilderFinished(Function<T, IStringTooltip> mapper) {
 			 * @return a {@link ListArgument}
 			 */
 			public ListArgument<T> buildGreedy() {
-				return new ListArgument<>(nodeName, delimiter, allowDuplicates, supplier, mapper);
+				return new ListArgument<>(nodeName, delimiter, allowDuplicates, supplier, mapper, skipListValidation);
```
DerEchtePilz, line 176: This option should not only be used here. If it should be added, the `ListTextArgument` should also be able to work with this option.

DerEchtePilz (changes requested):
A few things I noticed that I believe should be addressed.
Otherwise I guess I like this concept. It is a fairly niche feature but I don't see a reason why it shouldn't be added.

willkroboth:
Just a note that the command you implemented seems to have similar functionality to the `FlagsArgument` proposed by #483. Development on that is currently in progress on the `dev/command-build-rewrite` branch https://github.com/JorelAli/CommandAPI/commit/22c2ffac8b4ac9f3e0ac47f63cbe82c88729e44d.

Also, `dev/argument-exceptions` #476 currently [provides the API](https://github.com/JorelAli/CommandAPI/blob/8d40f573022487ab259adda74b8b6f9ec3980654/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java#L164-L222) for developers to manually suppress the exceptions thrown when the CommandAPI parses the ListArgument.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
stumper66: I was originally doing it this way but then discovered that the list was always empty so after some testing I decided to just have it return early instead.

DerEchtePilz: In this case, an alternate way of parsing this argument would have to be created. The list returned should contain the values provided by the user. I am not willing to say that users will have to keep track of their own list in order to validate results or something because returning an empty list is basically not acceptable.

DerEchtePilz:
This latest iteration doesn't work. I tested this command:
```java
new CommandAPICommand("test")
    .withArguments(new ListArgumentBuilder<String>("list")
        .skipListValidation(true)
        .withList("test")
         .withStringMapper()
        .buildGreedy()
    )
     .executesPlayer(info -> {
        List<String> list = (List<String>) info.args().get("list");
        info.sender().sendMessage(Component.text().content(list.toString()).build());
    })
    .register();
```
I then executed this command:
```
/test hello test police
```
The output I got was `[test]`, the expected result would be `[hello, test, police]` if the validation is skipped.

stumper66:
> This latest iteration doesn't work. I tested this command:
> 
> ```java
> new CommandAPICommand("test")
>     .withArguments(new ListArgumentBuilder<String>("list")
>         .skipListValidation(true)
>         .withList("test")
>          .withStringMapper()
>         .buildGreedy()
>     )
>      .executesPlayer(info -> {
>         List<String> list = (List<String>) info.args().get("list");
>         info.sender().sendMessage(Component.text().content(list.toString()).build());
>     })
>     .register();
> ```
> 
> I then executed this command:
> 
> ```
> /test hello test police
> ```
> 
> The output I got was `[test]`, the expected result would be `[hello, test, police]` if the validation is skipped.

Yeah it's not the most intuitive since the list basically is only good for suggestions.
Afterwards you have to use the rawArgMap and it returns everything in one string.

```yml
            .executes(CommandExecutor { sender, args -> processCmd(sender, args) })
           # ...
            .withOptionalArguments(
                ListArgumentBuilder<String>("values")
                    .skipListValidation(true)
                    .withList { info -> buildTabSuggestions(info) }
                    .withStringMapper()
                    .buildGreedy()

           val values = args.rawArgsMap["values"]
          val args = splitStringWithQuotes(values) # split into array and treat quotes as one element
```

Here's how I used it https://github.com/ArcanePlugins/LevelledMobs/blob/4.0.0-dev/levelledmobs-plugin/src/main/kotlin/io/github/arcaneplugins/levelledmobs/commands/subcommands/SummonSubcommand.kt#L56

DerEchtePilz:
Nope, that's not how it's going to work when merged. As I said, it would be best to have a separate parsing method when list validation is skipped.
Also, make sure to also update the `ListTextArgument` constructor so this also can work with this parameter.

stumper66:
I'll probably just cancel this PR and wait for `FlagsArgument` proposed by https://github.com/JorelAli/CommandAPI/issues/483 since it likely works better and is more intuitive.

DerEchtePilz:
> I'll probably just cancel this PR and wait for `FlagsArgument` proposed by #483 since it likely works better and is more intuitive.

I mean, do what you want to do, but again, this, despite being a really niche feature, can be useful. It just needs a bit more attention since there are expectations when using the list.

DerEchtePilz:
> Yeah it's not the most intuitive since the list basically is only good for suggestions.

Also, not true. The list, if validations are not skipped, provides the player or any other command executor with a list of _options_. The items in the suggestion list are there to tell the user the valid options.
The command sender can then choose which of these items should be used (compare the GIF in the ListArgument's documentation page, there the materials are suggested iirc).
The list that you then can get in the executor contains the chosen values. If not all values are selected, the list that provides suggestions and the list returned by the argument can be very different in size.

DerEchtePilz:
@stumper66 
Are you still working on this? If not, it might be better to close this PR since I don't think we want PRs sitting here that are not active anymore.

stumper66:
Cancelling the PR

PR #523, opened 2024-02-24
Exposed details of the CommandPermission

Timongcraft:
Made the optional string permission and negation state in the CommandPermission public

### DIFF
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java
@@ -125,16 +125,26 @@ public class CommandPermission {
 		return (negated ? "not " : "") + result;
 	}
 
+	/**
+	 * Returns the custom permission in string form if set
+	 * 
+	 * @return An {@code Optional<String>} with the custom permission or an empty {@code Optional} if not set
+	 */
 	public Optional<String> getPermission() {
 		return Optional.ofNullable(this.permission);
 	}
 
-	PermissionNode getPermissionNode() {
-		return this.permissionNode;
+	/**
+	 * Returns if the permission is negated
+	 * 
+	 * @return the permissions negation state
+	 */
+	public boolean isNegated() {
+		return this.negated;
 	}
 
-	boolean isNegated() {
-		return this.negated;
+	PermissionNode getPermissionNode() {
+		return this.permissionNode;
 	}
 
 	CommandPermission negate() {
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java
@@ -137,7 +137,7 @@ public class CommandPermission {
 	/**
 	 * Returns if the permission is negated
 	 * 
-	 * @return the permissions negation state
+	 * @return the permission's negation state
 	 */
 	public boolean isNegated() {
 		return this.negated;
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java
@@ -125,7 +125,7 @@ public class CommandPermission {
 		return (negated ? "not " : "") + result;
 	}
 
-	Optional<String> getPermission() {
+	public Optional<String> getPermission() {
 		return Optional.ofNullable(this.permission);
 	}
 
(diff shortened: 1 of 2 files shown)


willkroboth (commented):
Definitely works as expected.

Would it be useful to make these other methods in `CommandPermission` public?

https://github.com/JorelAli/CommandAPI/blob/46d1fa14990f9d3b6b9611e08f4e953f886291c9/commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java#L128-L143

I'm not sure if it is necessary to mention this method in the docs: https://commandapi.jorel.dev/9.3.0/permissions.html. Maybe at least add javadocs to the method now that it's public.

You could also add this change to the Changelog in the README:

https://github.com/JorelAli/CommandAPI/blob/46d1fa14990f9d3b6b9611e08f4e953f886291c9/README.md?plain=1#L418-L425

Timongcraft:
Probably don't, because the permissionNode can be null, and nullables should not be exposed according to the contribution guide and Skepter on Discord.
Also, does it make sense to update it in the 9.3.0 documentation since it is meant for 9.4.0? Should this even be documented?
And where would you want to add a Javadoc about this?

willkroboth:
> Probably don't, because the permissionNode can be null, and nullables should not be exposed according to the contribution guide and Skepter on Discord.

Makes sense. I suppose you could also change the method signature of `getPermissionNode` to be an Optional. It might be helpful in the case when you want to check if the permission is `OP` or `NONE`. Although, `CommandPermission` objects with a non-null `permissionNode` should only be these constant fields since the constructor is private. So, a developer could just check if their `CommandPermission` object is equal to one of these.

https://github.com/JorelAli/CommandAPI/blob/46d1fa14990f9d3b6b9611e08f4e953f886291c9/commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java#L44-L52

The `isNegated` method is probably important though for telling the difference between **has** `permissionA` and **does not have** `permissionA`

> Also, does it make sense to update it in the 9.3.0 documentation since it is meant for 9.4.0? Should this even be documented?

Yeah, you wouldn't change the old docs since the method wouldn't be public in 9.3.0. You'd change the source file here: https://github.com/JorelAli/CommandAPI/blob/46d1fa14990f9d3b6b9611e08f4e953f886291c9/docssrc/src/permissions.md. Then once the 9.4.0 docs are generated from the source files the change will be visible.


> And where would you want to add a Javadoc about this?

I meant a Javadoc on the `getPermission` method. Public members should probably all have Javadocs, though I'm sure there's some places where they don't :P.

Timongcraft:
> Makes sense. I suppose you could also change the method signature of `getPermissionNode` to be an Optional. It might be helpful in the case when you want to check if the permission is `OP` or `NONE`. Although, `CommandPermission` objects with a non-null `permissionNode` should only be these constant fields since the constructor is private. So, a developer could just check if their `CommandPermission` object is equal to one of these.
> 
I wouldn't change the signature of getPermissionNode because, as you mentioned, they could just check with the static instances. And for those who, for any reason, use reflection to access it, it remains accessible.
> 
> https://github.com/JorelAli/CommandAPI/blob/46d1fa14990f9d3b6b9611e08f4e953f886291c9/commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java#L44-L52
> 
> The `isNegated` method is probably important though for telling the difference between **has** `permissionA` and **does not have** `permissionA`
> 
Yea, makes sense.
> 
> Yeah, you wouldn't change the old docs since the method wouldn't be public in 9.3.0. You'd change the source file here: https://github.com/JorelAli/CommandAPI/blob/46d1fa14990f9d3b6b9611e08f4e953f886291c9/docssrc/src/permissions.md. Then once the 9.4.0 docs are generated from the source files the change will be visible.
> 
> But also yeah, I'm not sure if it's a big enough deal to mention?
> 
It doesn't really fit in the docs, so I would just not mention it there.
> 
> I meant a Javadoc on the `getPermission` method. Public members should probably all have Javadocs, though I'm sure there's some places where they don't :P.
> 
Makes sense. I thought you meant we should mention the change somewhere in the Javadocs.

willkroboth (approved):

commandapi-core/src/main/java/dev/jorel/commandapi/CommandPermission.java
```diff
@@ -125,16 +125,26 @@ public String toString() {
...
+	public Optional<String> getPermission() {
 		return Optional.ofNullable(this.permission);
 	}
 
-	PermissionNode getPermissionNode() {
-		return this.permissionNode;
+	/**
+	 * Returns if the permission is negated
+	 * 
+	 * @return the permissions negation state
```
JorelAli, line 140: "permissions" is missing an apostrophe.

Issue #536, opened 2024-04-02
Help for MultiLiteralArgument shows the arg name instead of the literal text


globau:
### CommandAPI version

9.3.0

### Minecraft version

1.20.4

### Are you shading the CommandAPI?

Yes

### What I did

```
    new CommandAPICommand("invite")
        .withShortDescription("Allow a player to join the server")
        .withArguments(new MultiLiteralArgument("type", "java", "bedrock"))
        .withArguments(new StringArgument("player"))
        .executes(
            (CommandSender sender, CommandArguments args) ->
                ...
```

### What actually happened

```
> help invite
[22:28:16 INFO]: --------- Help: /invite -------------------------
[22:28:16 INFO]: Allow a player to join the server
[22:28:16 INFO]: Usage:
[22:28:16 INFO]: - /invite <type> <player>
[22:28:16 INFO]: - /invite <type> <player>
```

### What should have happened

```
> help invite
[22:28:16 INFO]: --------- Help: /invite -------------------------
[22:28:16 INFO]: Allow a player to join the server
[22:28:16 INFO]: Usage:
[22:28:16 INFO]: - /invite java <player>
[22:28:16 INFO]: - /invite bedrock <player>
```

Ideally if the number of options is small (eg. <= 5) all options should be listed, otherwise just one line with the name:

```
> help invite
[22:28:16 INFO]: --------- Help: /invite -------------------------
[22:28:16 INFO]: Allow a player to join the server
[22:28:16 INFO]: Usage:
[22:28:16 INFO]: - /invite <type> <player>
```

DerEchtePilz:
Thanks for your report!

Unfortunately, this isn't as straight forward as I initially thought because at the point where the usage is generated, we do have no knowledge about which argument is used at which point.
I am interested in looking into this as for literal arguments the literal value should be displayed instead of the node name.

JorelAli:
Fixed in version 9.4.0.

PR #537, opened 2024-04-03
Updates help generation for literal arguments

DerEchtePilz:
For #536
Now displays the literal value instead of the node name

### DIFF
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
@@ -615,10 +615,12 @@ extends AbstractArgument<?, ?, Argument, CommandSender>
 		}
 
 		List<String> argumentsString = new ArrayList<>();
+		List<AbstractArgument<?, ?, ?, ?>> arguments = new ArrayList<>();
 		for (Argument arg : args) {
+			arguments.add(arg);
 			argumentsString.add(arg.getNodeName() + ":" + arg.getClass().getSimpleName());
 		}
-		RegisteredCommand registeredCommandInformation = new RegisteredCommand(commandName, argumentsString, shortDescription,
+		RegisteredCommand registeredCommandInformation = new RegisteredCommand(commandName, argumentsString, List.copyOf(arguments), shortDescription,
 			fullDescription, usageDescription, aliases, permission, namespace);
 		registeredCommands.add(registeredCommandInformation);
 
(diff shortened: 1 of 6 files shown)


commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIHandler.java
```diff
@@ -615,10 +615,12 @@ void register(CommandMetaData<CommandSender> meta, final Argument[] args,
 		}
 
 		List<String> argumentsString = new ArrayList<>();
+		List<AbstractArgument<?, ?, ?, ?>> arguments = new ArrayList<>();
 		for (Argument arg : args) {
+			arguments.add(arg);
 			argumentsString.add(arg.getNodeName() + ":" + arg.getClass().getSimpleName());
 		}
-		RegisteredCommand registeredCommandInformation = new RegisteredCommand(commandName, argumentsString, shortDescription,
+		RegisteredCommand registeredCommandInformation = new RegisteredCommand(commandName, argumentsString, List.copyOf(arguments), shortDescription,
```
JorelAli, line 623: This looks weird. We have an array of arguments `final Argument[] args`. We iterate over it to add it to a list, then we create an immutable copy of that list. Wouldn't it be better to just go straight to using `List.of(args)`?

commandapi-core/src/main/java/dev/jorel/commandapi/arguments/AbstractArgument.java
```diff
@@ -361,6 +361,10 @@ public void copyPermissionsAndRequirements(Argument argument) {
 		this.withPermission(argument.getArgumentPermission());
 	}
 
+	public String getAsArgumentString() {
```
JorelAli, line 364: Nitpick: This method name is confusing. This wraps the argument's _node name_ in `<` and `>` characters, so I would expect this method's name to mention something along the lines of _node name_.

JorelAli (changes requested):
Looks good - just a few minor changes please.

commandapi-core/src/main/java/dev/jorel/commandapi/arguments/AbstractArgument.java
DerEchtePilz: Hmm, I was thinking about this in a different way. The help generation originally used the `RegisteredCommand#argsAsStr()` method which contents will be something along the lines of `nodeName:ClassName`.
When the help is generated, this split at the colon and the node name is wrapped in `<` and `>`.
Hence I came up with that name.
I don't think a name along the lines of *node name* makes much sense, or maybe I currently can't think of a nice name.
If I had to change this name, I guess I'd currently go with `getHelpString` or `getAsArgumentHelpString` but those aren't *that* great either.
I do agree that the current name isn't fitting now that you mention it and I could think about this again but I just don't have a good alternative at the moment.

JorelAli: I think `getHelpString` works - it's the string representation of an argument for use in command help.

JorelAli (approved):

Issue #535, opened 2024-03-31
Cast object to class like as it is done in JOOQ


Hxncusik:
### Description

Record#get(String var1, Class<? extends U> var2)
For example:
String item = entry.get("item", String.class);
double chance = entry.get("chance", Double.class);
double range = entry.get("range", Double.class);

### Expected code

new CommandAPICommand("test").withPermission("*")
                .withArguments(new FloatArgument("chance"), new FloatArgument("range"))
                .executesPlayer((player, args) -> {
                    float chance = args.get("chance", float.class);
                    float range = args.get("range", float.class);
                });

DerEchtePilz:
First off, I like this idea and this can actually be grouped with a discussion that took place on the CommandAPI discord server about two weeks ago where the general gist was to use argument instances to safe-cast arguments.
I also planned working on it eventually so thanks for your suggestion as I believe Brigadier does this like this too, I will definitely consider this when coming up with a potential solution.

Hxncusik:
Thank you for your reply, I'm very pleased to hear that. I will look forward to implementing the new features.

JorelAli:
Implemented in version 9.4.0.

PR #540, opened 2024-04-06
Implement "safe-casting"

DerEchtePilz:
This PR is supposed to implement certain methods as suggested by Hawk [here](https://discord.com/channels/745416925924032523/745419648970784821/1218161950865883198) and by #535.

Currently, there is a bit more testing I want to do, additionally I am not sure if this would actually be something we want.
If so, there is a bit more work required:
- [x] Global changelog entry
- [x] Update [CommandArguments](https://commandapi.jorel.dev/9.3.0/commandarguments.html), [Documentation changelog](https://commandapi.jorel.dev/9.3.0/intro.html#documentation-changelog)

### DIFF
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
@@ -6,6 +6,7 @@ import javax.annotation.Nullable;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
@@ -47,6 +48,17 @@ public record CommandArguments(
 	 */
 	String fullInput
 ) {
+
+	private static final Map<Class<?>, Class<?>> PRIMITIVE_TO_WRAPPER = new HashMap<>() {{
+		put(boolean.class, Boolean.class);
+		put(char.class, Character.class);
+		put(byte.class, Byte.class);
+		put(short.class, Short.class);
+		put(int.class, Integer.class);
+		put(long.class, Long.class);
+		put(float.class, Float.class);
+		put(double.class, Double.class);
+	}};
 
 	// Access the inner structure directly
 
@@ -487,10 +499,10 @@ public record CommandArguments(
 		if (argument == null) {
 			return null;
 		}
-		if (!argument.getClass().equals(argumentType)) {
+		if (!argument.getClass().equals(PRIMITIVE_TO_WRAPPER.getOrDefault(argumentType, argumentType))) {
 			return null;
 		}
-		if (!argumentType.isAssignableFrom(argument.getClass())) {
+		if (!PRIMITIVE_TO_WRAPPER.getOrDefault(argumentType, argumentType).isAssignableFrom(argument.getClass())) {
 			return null;
 		}
 		return (T) argument;
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
@@ -466,7 +466,7 @@ public record CommandArguments(
 	@Nullable
 	public <T> T getByArgument(AbstractArgument<T, ?, ?, ?> argumentType) {
 		Object argument = get(argumentType.getNodeName());
-		return castArgument(argument, argumentType.getPrimitiveType());
+		return castArgument(argument, argumentType.getPrimitiveType(), argumentType.getNodeName());
 	}
 
 	/**
@@ -479,7 +479,7 @@ public record CommandArguments(
 	@Nullable
 	public <T> T getByClass(String nodeName, Class<T> argumentType) {
 		Object argument = get(nodeName);
-		return castArgument(argument, argumentType);
+		return castArgument(argument, argumentType, nodeName);
 	}
 
 	/**
@@ -492,18 +492,36 @@ public record CommandArguments(
 	@Nullable
 	public <T> T getByClass(int index, Class<T> argumentType) {
 		Object argument = get(index);
-		return castArgument(argument, argumentType);
+		// This seems like an extremely bogus implementation but I really don't want
+		// to figure the node name vs index stuff out in the castArgument method
+		int argumentIndex = 0;
+		for (String nodeName : argsMap.keySet()) {
+			if (argumentIndex != index) {
+				argumentIndex++;
+				continue;
+			}
+			// Ensure the found node name points to the right argument
+			if (argument == null) {
+				return null;
+			}
+			if (!argument.equals(argsMap.get(nodeName))) {
+				throw new IllegalStateException("Unexpected behaviour detected while retrieving argument (arguments don't match up)!" +
+					"This should never happen - if you're seeing this message, please" +
+					"contact the developers of the CommandAPI, we'd love to know how you managed to get this error!");
+			}
+			return castArgument(argument, argumentType, nodeName);
+		}
+		throw new IllegalStateException("Unexpected behaviour detected while retrieving argument (didn't find the node name)!" +
+			"This should never happen - if you're seeing this message, please" +
+			"contact the developers of the CommandAPI, we'd love to know how you managed to get this error!");
 	}
 
-	private <T> T castArgument(Object argument, Class<T> argumentType) {
+	private <T> T castArgument(Object argument, Class<T> argumentType, String name) {
 		if (argument == null) {
 			return null;
 		}
-		if (!argument.getClass().equals(PRIMITIVE_TO_WRAPPER.getOrDefault(argumentType, argumentType))) {
-			return null;
-		}
 		if (!PRIMITIVE_TO_WRAPPER.getOrDefault(argumentType, argumentType).isAssignableFrom(argument.getClass())) {
-			return null;
+			throw new IllegalArgumentException("Argument '" + name + "' is defined as " + argument.getClass().getSimpleName() + ", not " + argumentType);
 		}
 		return (T) argument;
 	}
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
@@ -449,7 +449,7 @@ public record CommandArguments(
 	 * Returns an argument purely based on its CommandAPI representation. This also attempts to directly cast the argument to the type represented by {@link dev.jorel.commandapi.arguments.AbstractArgument#getPrimitiveType()}
 	 *
 	 * @param argumentType The argument instance used to create the argument
-	 * @return The argument represented by the CommandAPI argument, or null if the argument's clas cannot be cast to the type represented by {@link dev.jorel.commandapi.arguments.AbstractArgument#getPrimitiveType()}
+	 * @return The argument represented by the CommandAPI argument, or null if the argument's class cannot be cast to the type represented by {@link dev.jorel.commandapi.arguments.AbstractArgument#getPrimitiveType()}
 	 */
 	@Nullable
 	public <T> T getByArgument(AbstractArgument<T, ?, ?, ?> argumentType) {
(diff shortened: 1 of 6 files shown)


commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -443,4 +445,55 @@ public <T> Optional<T> getOptionalUnchecked(String nodeName) {
...
+	private <T> T castArgument(Object argument, Class<T> argumentType) {
+		if (argument == null) {
+			return null;
+		}
+		if (!argument.getClass().equals(argumentType)) {
+			return null;
+		}
+		if (!argumentType.isAssignableFrom(argument.getClass())) {
+			return null;
+		}
```
JorelAli, line 495: I might be missing a few things, but I think this doesn't work as expected - surely we want the `isAssignableFrom` check first?

Secondly, I don't think we should be returning `null` here (in the case that the classes don't match). Surely:

- We know what class the user provided for `argumentType`
- We know what class the actual `argument` is

Therefore, we can throw a runtime error stating something along the lines of "Expected <argument name or index> to be a <argument type>, but was given a <user provided class> instead".

DerEchtePilz: Interesting thoughts, as I said, I definitely want to run some tests next.
This whole thing was more or less inspired by [CommandContext#getArgument(String, Class)](https://github.com/Mojang/brigadier/blob/master/src/main/java/com/mojang/brigadier/context/CommandContext.java#L97-L111)
This is why I put the null check first but yeah, I guess we can flip the last two statements.

As for the `null`'s, I thought it was the better choice since we can have it return `null` anyway. If we  think exceptions are better, I certainly can change this.

Brigadier also uses those wrapper classes for the class check, maybe this is also something I have to look into. Again, I definitely will run some tests tomorrow, let's see what I get from those.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -443,4 +445,55 @@ public <T> Optional<T> getOptionalUnchecked(String nodeName) {
 		return (Optional<T>) getOptional(nodeName);
 	}
 
+	/**
+	 * Returns an argument purely based on its CommandAPI representation. This also attempts to directly cast the argument to the type represented by {@link dev.jorel.commandapi.arguments.AbstractArgument#getPrimitiveType()}
+	 *
+	 * @param argumentType The argument instance used to create the argument
+	 * @return The argument represented by the CommandAPI argument, or null if the argument's clas cannot be cast to the type represented by {@link dev.jorel.commandapi.arguments.AbstractArgument#getPrimitiveType()}
```
XHawk87, line 452: Typo: `clas`. Minor thing, but that's going to sit there forever otherwise :smile:

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
XHawk87: Can we include support for optionals in this too, or would you prefer a separate PR for that?

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
DerEchtePilz: Thanks xD

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
DerEchtePilz: > Can we include support for optionals in this too, or would you prefer a separate PR for that?

We can do anything. If you'd like optionals, I guess I could arrange adding a method returning an optional too. And I don't think a additional PR is necessary, this is supposed to add this.

DerEchtePilz:
Someone please review the latest changes, not sure if we want that behaviour.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
DerEchtePilz: > Can we include support for optionals in this too, or would you prefer a separate PR for that?

Methods for `Optional`'s (and default values for consistency) have been added 👍

willkroboth:
Just a note that some arguments use a raw type as their `T` generic parameter. For example, `MapArgument<K, V>` extends `Argument<LinkedHashMap>` rather than `Argument<LinkedHashMap<K, V>>`

https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/MapArgument.java#L26

If `MapArgument` did extend `Argument<LinkedHashMap<K, V>>`, then the `getPrimitiveType` method would have to return a `Class<LinkedHashMap<K, V>>` object, which Java doesn't like very much: https://stackoverflow.com/questions/1079279/class-object-of-generic-class-java.

That just means this code has a little warning:

```java
MapArgument<String, Integer> argument = new MapArgumentBuilder<String, Integer>("map")
	.withKeyMapper(s -> s)
	.withValueMapper(s -> Integer.parseInt(s))
	.withoutKeyList()
	.withoutValueList()
	.build();

new CommandAPICommand("command")
	.withArguments(argument)
	.executes(info -> {
		// Type safety: The expression of type LinkedHashMap needs unchecked conversion to conform to Map<String,Integer>
		Map<String, Integer> map = info.args().getByArgument(argument);
		info.sender().sendMessage(map.toString());
	})
	.register();
```

This note might also be relevant for https://github.com/JorelAli/CommandAPI/pull/545. It's not a big problem, but the two systems don't currently work perfectly for these arguments:

- [`AxisArgument`](https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/AxisArgument.java#L39)
- [`BlockPredicateArgument`](https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/BlockPredicateArgument.java#L39)
- [`EntitySelectorArgument.ManyEntities`](https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/EntitySelectorArgument.java#L125)
- [`EntitySelectorArgument.ManyPlayers`](https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/EntitySelectorArgument.java#L169)
- [`ItemStackPredicateArgument`](https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ItemStackPredicateArgument.java#L37)
- `ListArgument` and `ListTextArgument` via [`ListArgumentCommon`](https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java#L35)
- [`MapArgument`](https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/MapArgument.java#L26)
- [`ParticleArgument`](https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ParticleArgument.java#L35)
- [`ScoreHolderArgument.Multiple`](https://github.com/JorelAli/CommandAPI/blob/a7d2717450c017002647fe611a3d39deda6874d2/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ScoreHolderArgument.java#L79)

DerEchtePilz:
True, but if I am not mistaken, this is the best we can do here.
Since it additionally doesn't work for the `CustomArgument` anyway, I guess in terms of safety we can ignore the generic types.

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -443,4 +457,153 @@ public <T> Optional<T> getOptionalUnchecked(String nodeName) {
...
+		return Optional.ofNullable(argument);
+	}
+
+	private <T> T castArgument(Object argument, Class<T> argumentType, Object argumentNameOrIndex) {
+		if (argument == null) {
+			return null;
+		}
+		if (!PRIMITIVE_TO_WRAPPER.getOrDefault(argumentType, argumentType).isAssignableFrom(argument.getClass())) {
+			throw new IllegalArgumentException(buildExceptionMessage(argumentNameOrIndex, argument.getClass().getSimpleName(), argumentType.getSimpleName()));
+		}
```
JorelAli, line 593: I don't get what's going on here. Say I'm using a command with a single `AngleArgument` argument (which returns a `float` primitive).

Say we use the following to get the first argument:
```java
getByClass(0, float.class);
```

Using `PRIMITIVE_TO_WRAPPER` gives us:
```java
!PRIMITIVE_TO_WRAPPER.getOrDefault(float.class, float.class).isAssignableFrom(float.class)
```
which evaluates to the following:
```java
!Float.class.isAssignableFrom(float.class)
```
which evaluates to:
```java
!false
```
So we always throw an exception here? Or is my static-code analysis deceiving me and I'm missing something very obvious here?

JorelAli (changes requested):
A couple of minor documentation changes and code linting tweaks (avoid the use of variables which declared non-final and are only ever used once). A critical change required for `PRIMITIVE_TO_WRAPPER`

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
DerEchtePilz: Yeah, this is one of first things I tested.
I think what's going on here is that because we store every argument as an `Object`, the primitive types are being boxed which is why we're not comparing to the primitive types.
So the `AngleArgument` becomes something like this:
```java
getByClass(0, float.class);
```
Then the primitive wrapper thing:
```java
!PRIMITIVE_TO_WRAPPER.getOrDefault(float.class, float.class).isAssignableFrom(Float.class)
```

This evaluates to:
```java
!Float.class.isAssignableFrom(Float.class)
```

This finally evaluates to:
```java
!true
```

commandapi-core/src/main/java/dev/jorel/commandapi/executors/CommandArguments.java
```diff
@@ -46,6 +49,17 @@ public record CommandArguments(
...
+	private static final Map<Class<?>, Class<?>> PRIMITIVE_TO_WRAPPER = new HashMap<>() {{
+		put(boolean.class, Boolean.class);
+		put(char.class, Character.class);
+		put(byte.class, Byte.class);
+		put(short.class, Short.class);
+		put(int.class, Integer.class);
+		put(long.class, Long.class);
+		put(float.class, Float.class);
+		put(double.class, Double.class);
+	}};
```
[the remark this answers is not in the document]
DerEchtePilz: Sure, this already done locally, but may I ask why?

JorelAli (approved):

Issue #544, opened 2024-04-13
Type-safe Kotlin DSL


sya-ri:
### Description

The current Kotlin DSL requires casts, so it is not type-safe. We can't get the most out of Kotlin.

> ```kotlin
> commandTree("sendmessageto") {
>     playerArgument("player") {
>         greedyStringArgument("msg") {
>             anyExecutor { _, args ->
>                 val player: Player = args["player"] as Player
>                 val message: String = args["msg"] as String
>                 player.sendMessage(message)
>             }
>         }
>     }
> }
> ```
> 
> ```kotlin
> commandAPICommand("sendmessageto") {
>     playerArgument("player")
>     greedyStringArgument("msg")
>     anyExecutor { _, args ->
>         val player: Player = args["player"] as Player
>         val message: String = args["msg"] as String
>         player.sendMessage(message)
>     }
> }
> ```
> from: [Defining a simple message command](https://commandapi.jorel.dev/9.3.0/kotlindsl.html#defining-a-simple-message-command)

### Expected code

```kotlin
commandTree("sendmessageto") {
    playerArgument("player") { getPlayer ->
        greedyStringArgument("msg") { getMessage ->
            anyExecutor { _, args ->
                val player = getPlayer(args) // Returns Player
                val message = getMessage(args) // Returns String
                player.sendMessage(message)
            }
        }
    }
}
```

I have confirmed that this code can actually be rewritten. You can try it out using this library: https://github.com/sya-ri/commandapi-kotlin-improved

In this library, the Kotlin DSL for CommandAPICommand is not supported, but we should be able to rewrite it similarly.

### Extra details

If necessary, I would like to submit a PR for the changes made by my library 😉

DerEchtePilz:
Hmm, interesting.
If you choose to open a PR, please note that this should be an additional option, not a replacement.
Also, this should also be supported in the `commandAPICommand` structure without the need to redefine how that looks.

sya-ri:
@DerEchtePilz Thank you for your interest.

We will need to make a breaking change to "optional".

Currently, specifying `optional = true` for argument makes it **nullable**, and specifying false makes it not.

For example:

```kotlin
commandTree("sendmessageto") {
    playerArgument("player") { getPlayer ->
        greedyStringArgument("msg", optional = true) { getMessage ->
            anyExecutor { _, args ->
                val player = getPlayer(args) // Expected: Player (not-null)
                val message = getMessage(args) // Expected: String? (nullable)
[...]

PR #545, opened 2024-04-16
Implement Type-safe Kotlin DSL

sya-ri:
#544 

---

I didn't change the documentation because I didn't really understand it.

### DIFF
--- a/commandapi-kotlin/commandapi-bukkit-kotlin/src/main/kotlin/dev/jorel/commandapi/kotlindsl/CommandAPICommandDSL.kt
+++ b/commandapi-kotlin/commandapi-bukkit-kotlin/src/main/kotlin/dev/jorel/commandapi/kotlindsl/CommandAPICommandDSL.kt
@@ -2,10 +2,31 @@
 
 package dev.jorel.commandapi.kotlindsl
 
-import dev.jorel.commandapi.*
+import dev.jorel.commandapi.CommandAPICommand
 import dev.jorel.commandapi.arguments.*
+import dev.jorel.commandapi.wrappers.*
+import dev.jorel.commandapi.wrappers.Rotation
+import net.kyori.adventure.text.Component
+import net.kyori.adventure.text.format.TextColor
+import net.md_5.bungee.api.chat.BaseComponent
+import org.bukkit.*
+import org.bukkit.advancement.Advancement
+import org.bukkit.block.Biome
+import org.bukkit.block.Block
+import org.bukkit.block.data.BlockData
 import org.bukkit.command.CommandSender
+import org.bukkit.enchantments.Enchantment
+import org.bukkit.entity.Entity
+import org.bukkit.entity.EntityType
+import org.bukkit.entity.Player
+import org.bukkit.inventory.ItemStack
+import org.bukkit.inventory.Recipe
+import org.bukkit.loot.LootTable
 import org.bukkit.plugin.java.JavaPlugin
+import org.bukkit.potion.PotionEffectType
+import org.bukkit.scoreboard.Objective
+import org.bukkit.scoreboard.Team
+import java.util.*
 import java.util.function.Predicate
 
 inline fun commandAPICommand(name: String, command: CommandAPICommand.() -> Unit = {}) = CommandAPICommand(name).apply(command).register()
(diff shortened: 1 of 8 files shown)


DerEchtePilz (changes requested):
Again, this should not be a replacement. This should rather be an alternative for those who want this.
Please also remove the additional methods containing the `Optional` part, especially in the `CommandTreeDSL` because as mentioned in your issue, optional arguments are going to be removed from `CommandTree`s in the future. Regardless, these methods are not wanted in the `CommandAPICommandDSL` either.

Other than that I honestly don't really understand how this works exactly so I definitely want to test this myself first, but that's obviously my problem :D

DerEchtePilz:
> I didn't change the documentation because I didn't really understand it.

The documentation is built using mdbook, I believe there are download links for the required executables in the wiki here on GitHub.
The source for the documentation consists of `.md` files. Those can be found in `docssrc/src`.
The documentation additionally uses compiled code so the code examples can be used in code without paying attention to the correct syntax. Its source resides in the `commandapi-documentation-code`. This module is the Bukkit module, there is not really documentation for Velocity yet.

sya-ri:
> Again, this should not be a replacement. This should rather be an alternative for those who want this.

Does that mean I should leave the method's arguments and processing as they are rather than changing them, and define a new method in another file?

The only incompatibility caused by this replacement is when specifying `argument(..., optional = true)` or `argument(..., optional = false)`; you should remove `optional = false` or use `optionalArgument(...)`.

Retrieval methods using get and cast are still supported. I just made it possible to use the getter inside the lambda.

https://github.com/JorelAli/CommandAPI/pull/545/files#diff-a396609c4f42dbc82e492e9976de00ceb4c97805935fc741902fbb6cb4eae000
https://github.com/JorelAli/CommandAPI/pull/545/files#diff-67d7c60b582802c1efad0034b5b7bf5da58e284935c795f2959d5aa5bab2e654
https://github.com/JorelAli/CommandAPI/pull/545/files#diff-db52ae2a394c8abc971e1c4845fb833df06243d1096b6e8938dd042ed4eead99

I only changed the optional tests and added some tests.

DerEchtePilz:
> Does that mean I should leave the method's arguments and processing as they are rather than changing them, and define new methods in another file?

Yeah, that would be ideal.

> The only incompatibility caused by this replacement is when specifying `argument(..., optional = false)` or `argument(..., optional = true);` you should remove `optional = false` or use `optionalArgument(...).`

I mean, technically you can do whatever here, just don't introduce duplicate methods like the ones that have the `Optional` in their name. That is something we definitely don't want.

DerEchtePilz:
Hmm, I just realized, we do have [Delegated properties](https://commandapi.jorel.dev/9.3.0/delegated_properties.html) support for the Kotlin DSL.

That of course doesn't prevent this feature from being added but generally it is sort of the same functionality, right?

willkroboth:
I think delegated properties are still syntactic sugar for `CommandArguments#getUnchecked`. You could still accidentally put the wrong class, and you wouldn't know until runtime when there's a `ClassCastException` or something.

```kotlin
commandTree("sendmessageto") {
    playerArgument("player") {
        greedyStringArgument("msg") {
            anyExecutor { _, args ->
                val player: String by args // At runtime: Inconvertible types, Player is not a String
                val message: String by args
                player.sendMessage(message)
            }
        }
    }
}
```

The changes here allow there to be a compile time exception. IDEs can also statically analyze the code and catch the error, which makes it easier for the developer to notice and fix.

```kotlin
commandTree("sendmessageto") {
    playerArgument("player") { getPlayer ->
        greedyStringArgument("msg") { getMessage ->
            anyExecutor { _, args ->
                String player = getPlayer(args) // At compile time: Inconvertible types, Player is not a String
                val message = getMessage(args)
                player.sendMessage(message)
            }
        }
    }
}
```

DerEchtePilz:
Yeah, true. I mean, it also depends on the variable name you put there so the type safety is still only sort of there.
I mean yeah, in the end a call to `CommandArguments#get(String) as T` is done because the unsafe `CommandArguments#getUnchecked` wouldn't exactly work here (as we don't exactly work with fixed types there I think)

Issue #538, opened 2024-04-03
Help generation doesn't work for the namespaced version of a command


DerEchtePilz:
### CommandAPI version

9.4.0-SNAPSHOT

### Minecraft version

1.20.4

### Are you shading the CommandAPI?

Yes

### What I did

I encountered this bug while testing the fix for #536 . I have my `CommandAPIBukkitConfig` set up as follows:
```java
CommandAPI.onLoad(new CommandAPIBukkitConfig(this)
    .missingExecutorImplementationMessage("You cannot execute this command!")
    .dispatcherFile(new File(getDataFolder(), "command_registration.json"))
    .usePluginNamespace()
);
```

I then registered this command:
```kt
commandAPICommand("test") {
    multiLiteralArgument(nodeName = "type", "java", "bedrock")
    stringArgument("additional")
    playerExecutor { player, args ->
        val type: String by args
        player.sendMessage(Component.text().content(type))
    }
}
```

### What actually happened

I believe this screenshot shows it best:
![image](https://github.com/JorelAli/CommandAPI/assets/81232921/3395adcf-860e-4525-9e22-5497927b3c7f)

### What should have happened

The help menus should be identical.

### Other

This issue most likely isn't applicable to the latest snapshot version only. I believe it definitely happened before too.

JorelAli:
Fixed in version 9.4.0.

PR #546, opened 2024-04-22
Fix help topic generation

DerEchtePilz:
The basic thing works, just a few more things need checking:
- [x] ~~Figure out if this has been introduced by #509 or if this was a thing before. I did not check this when opening #538~~ Technically, namespaces having a help entry might actually be considered a bug since a command with the `minecraft:` namespace didn't have a help entry before the current snapshot phase.
- [x] ~~Alias generation currently is stupid for namespaced versions of a command~~ Nope, it isn't. See above.
- [x] ~~Should the usage generation be updated again to display the namespaced version of a command?~~ Not necessary, the namespaced version having a help topic is a bug introduced by #509

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
@@ -299,6 +299,10 @@ public abstract class CommandAPIBukkit<Source> implements CommandAPIPlatform<Arg
 		return (Bukkit.getPluginCommand(command) == null ? "/" : "/minecraft:") + command;
 	}
 
+	private String generateCommandHelpPrefix(String command, String namespace) {
+		return (Bukkit.getPluginCommand(command) == null ? "/" + namespace + ":" : "/minecraft:") + command;
+	}
+
 	private void generateHelpUsage(StringBuilder sb, RegisteredCommand command) {
 		// Generate usages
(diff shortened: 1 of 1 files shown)


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
```diff
@@ -354,10 +358,15 @@ private String[] getUsageList(RegisteredCommand currentCommand) {
 
 	void updateHelpForCommands(List<RegisteredCommand> commands) {
 		Map<String, HelpTopic> helpTopicsToAdd = new HashMap<>();
+		List<String> namespacedCommandNames = new ArrayList<>();
```
JorelAli, line 361: Order doesn't matter in this context, we could be using a `Set<>` instead.

JorelAli (approved):

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
DerEchtePilz: Alright, guess I'll change that real quick and then I assume I am good to merge this?

JorelAli: Yup, you're good to merge this to `dev/dev`.

PR #555, opened 2024-05-12
Hotfix CommandAPI throwing errors because of Paper's Brigadier command API

DerEchtePilz:
Some of this is not pretty but doing this the VarHandle-less way also got rid of the exception in the console which I didn't seem to do with VarHandles.

Nevertheless, this should make it possible to at least register commands using Paper 1.20.6 build 65 and onwards.

I plan to handle all this properly in #517


Sytm:
I gave this PR a quick spin with my plugin and it seems to work without any issues

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-nms-common/src/main/java/dev/jorel/commandapi/nms/NMS_Common.java
```diff
@@ -92,6 +94,18 @@
...
+
+	static {
+		Field temporary;
+		try {
+			temporary = MinecraftServer.class.getDeclaredField("vanillaCommandDispatcher");
+		} catch (Exception e) {
+			temporary = null;
+		}
+		commandDispatcher = temporary;
+	}
```
JorelAli, line 107: A field that could be null that doesn't have `@Nullable` or wrapped in an `Optional`! 🫢

PR #556, opened 2024-05-13
Change vanillaCommandDispatcher to getCommands()

toxicity188:
In current paper build, vanillaCommandDispatcher was removed. so it requires to use getCommands().


DerEchtePilz:
This is already hotfixed in 9.5.0-SNAPSHOT

PR #560, opened 2024-06-03
Delegate Bukkit command registering logic to `CommandRegistrationStrategy`

willkroboth:
This PR moves a lot of the current logic for registering commands on Bukkit into the class `SpigotCommandRegistration`. This allows an alternate implementation of the new `CommandRegistrationStrategy` interface to be loaded when internal changes from https://github.com/PaperMC/Paper/pull/8235 (Paper-1.20.6-65 and later) are present. The `CommandRegistrationStrategy` is created by the method `NMS#createCommandRegistrationStrategy`.

This is a more permanent fix than https://github.com/JorelAli/CommandAPI/pull/555 and resolves these issues when running on Paper-1.20.65 and later:
- Running `/minecraft:reload` deletes CommandAPI commands (Thanks FireML on [Discord](https://discord.com/channels/745416925924032523/745757702608912556/1245385777056190534))
- `unregisterBukkit` parameter of `CommandAPIBukkit#unregister` isn't fully taken into account

The implementation can still probably be refined by https://github.com/JorelAli/CommandAPI/pull/517 (e.g. CommandAPI-Spigot should never need `PaperCommandRegistration`, and later Paper versions should never need `SpigotCommandReigistration`)

The NMS methods `getResourcesDispatcher`, `isVanillaCommandWrapper`, `wrapToVanillaCommandWrapper`, and `isBukkitCommandWrapper` were moved to `SpigotCommandRegistration` because they only make sense if Paper changes are not present. These methods are only used in the CommandAPI by `SpigotCommandRegistration` anyway, but developers can still access them using `CommandAPIBukkit#getCommandRegistrationStrategy`. `PaperCommandRegistration` provides a similar method: `isBukkitCommand`.

### TODO:

Before merging I want to double check that the following works

- Commands register during server start up
- Commands register after server enables
- `unregisterBukkit` parameter is repsected
- Commands persist after `minecraft:reload`

On the following versions:

- [x] Latest Paper (Paper-1.20.6-131)
- [x] Paper-1.20.6-65
- [x] Paper-1.20.6-64
- [x] Paper-1.20.5
- [x] Paper-1.20.4

- [x] Spigot 1.20.6
- [x] Spigot 1.20.4

And of course~ code review is greatly appreciated!

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
@@ -1,0 +1,54 @@
+package dev.jorel.commandapi;
+
+import com.mojang.brigadier.CommandDispatcher;
+import com.mojang.brigadier.builder.LiteralArgumentBuilder;
+import com.mojang.brigadier.tree.LiteralCommandNode;
+
+import java.util.List;
+
+/**
+ * Handles logic for registering commands after Paper build 65, where https://github.com/PaperMC/Paper/pull/8235
+ * changed a bunch of the behind-the-scenes logic.
+ */
+public class PaperCommandRegistration<Source> extends CommandRegistrationStrategy<Source> {
+	// References to necessary objects
+	private final CommandDispatcher<Source> brigadierDispatcher;
+
+	public PaperCommandRegistration(CommandDispatcher<Source> brigadierDispatcher) {
+		this.brigadierDispatcher = brigadierDispatcher;
+	}
+
+	@Override
+	public CommandDispatcher<Source> getBrigadierDispatcher() {
+		return brigadierDispatcher;
+	}
+
+	@Override
+	public void runTasksAfterServerStart() {
+		// Nothing to do
+	}
+
+	@Override
+	public void postCommandRegistration(RegisteredCommand registeredCommand, LiteralCommandNode<Source> resultantNode, List<LiteralCommandNode<Source>> aliasNodes) {
+		// Nothing to do
+	}
+
+	@Override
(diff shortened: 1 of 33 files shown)


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -0,0 +1,99 @@
...
+	// References to necessary methods
+	private final Supplier<CommandDispatcher<Source>> getBrigadierDispatcher;
+	private final Predicate<CommandNode<Source>> isBukkitCommand;
+
+	// Store registered commands nodes for eventual reloads
+	private final RootCommandNode<Source> registeredNodes = new RootCommandNode<>();
+
+	public PaperCommandRegistration(
+		Supplier<CommandDispatcher<Source>> getBrigadierDispatcher, Predicate<CommandNode<Source>> isBukkitCommand
+	) {
```
DerEchtePilz, line 27: This is some me kind of nitpick but can we have this on one line?

DerEchtePilz (commented):
What's our standpoint on those `*` imports?
Do we want them? I personally tried to avoid them in general but if we also leave them like they are if we think they're fine.

JorelAli:
> What's our standpoint on those `*` imports? Do we want them?

We don't care. They don't change the output of the compiled code, the only downside to them is it's harder for humans to statically analyse code by staring at it in something other than an IDE, but it's 2024 and that's basically irrelevant.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
willkroboth: Fair enough. I think I expected to need to add more parameters (like SpigotCommandRegistration), but just two is a reasonable length.

DerEchtePilz (approved):
LGTM

PR #570, opened 2024-06-25
v9.5.1 mod

stumper66:
Added a new option: skipListValidation


DerEchtePilz:
So, I definitely would like to run some manual tests before reviewing, approving or merging anything, but it definitely looks good after initially looking over it.

DerEchtePilz:
Alright, I just ran some manual tests and I feel like there are some points that should be addressed/clarified as I think it's unexpected behaviour.
I registered this command: https://pastes.dev/f7lH64VJih
While the argument works as expected when not skipping anything, I did notice weird behaviour when using the argument where list validation is skipped:
```
/listtest skip a c a d b c e a
```

Executing this command led to this output:
```
[a, c, b, d]
```

This leads to the following questions:
- Is that behaviour correct? 
- So should the `allowDuplicates(false)` option (which is the default too) lead to only the error not appearing and duplicates still not being included? 
- Or should this option make it so that `allowDuplicates` is set to `true` and disable any further call to `allowDuplicates(boolean)` by checking if the `skipListValidation(boolean)` was used?

JorelAli:
> * Is that behaviour correct?

This doesn't look correct. I don't think `skipListValidation` and `allowDuplicates` are mutually exclusive - I think there's a case where you may want to skip list validation (allow any value in a list), but still explicitly want to prevent duplicates. That said, there are some unexpected results:

The user provides the list `a c a d b c e a`. As such, I would expect the result to be the following:

```
[a, c, d, b, e]
```

I would expect the order to be retained (I shouldn't see `[a, c, b, ...]`) and I would expect to see `e`, despite

> * So should the `allowDuplicates(false)` option (which is the default too) lead to only the error not appearing and duplicates still not being included?

For all intents and purposes, if `skipListValidation` is set, errors should not appear. If `allowDuplicates(false)` (which is by default) is set, duplicates should still be prevented.

-----

I think the naming of this feature leads to ambiguity (especially with duplicates) and I don't particularly like it. My understanding of this feature is it allows the `ListArgument` to permit any value which is not declared in the `withList()` method. This is a feature the CommandAPI does not have at the moment. Effectively, this is something more like a `.allowAnyValue(boolean)`

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentBuilder.java
```diff
@@ -56,6 +57,17 @@ public ListArgumentBuilder<T> allowDuplicates(boolean allowDuplicates) {
 		return this;
 	}
 
+	/**
+	 * Whether the list verification should be skipped
+	 *
+	 * @param allowAnyValue whether to skip list validation
```
JorelAli, line 63: Can we improve these JavaDocs?

DerEchtePilz:
> If `allowDuplicates(false)` (which is by default) is set, duplicates should still be prevented.

Should that still produce the error message if `allowAnyValue(true)` is set? That currently happens and I am not sure if it should.

-----

That being asked, there are still some problems that arise from at least one check not being bypassed.
I run the command:
```
/listtest skip a c e f b d w r t
```
and I get the output:
```
[a, c, b, d]
```

This is unexpected as the output should include every given letter in my command.

JorelAli:
> Should that still produce the error message if `allowAnyValue(true)` is set? That currently happens and I am not sure if it should.

This appears to be fixed in the latest commit.

DerEchtePilz:
> This appears to be fixed in the latest commit.

Well no, I tested the latest commit, that check apparently was removed.

stumper66:
I'm not sure why some of the input is missing. Debugging now.

DerEchtePilz:
> Should that still produce the error message if `allowAnyValue(true)` is set? That currently happens and I am not sure if it should.

After some discussion in discord (https://discord.com/channels/745416925924032523/745419648970784821/1255545195051618304), I must correct myself here, if duplicates are not allowed, the error should be displayed, the current behaviour (present in the latest commit) in that regard is correct.

stumper66:
I've messed up this PR enough, canceling it and creating a new one

PR #571, opened 2024-06-26
Add allowAnyValue

stumper66:
Added a new option: allowAnyValue

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
@@ -132,16 +132,18 @@ public class ListArgumentCommon<T> extends Argument<List> {
 	
 		// If the argument's value is in the list of values, include it
 		List<T> list = new ArrayList<>();
+		Set<String> listKeys = new HashSet<>(); // Set to keep track of duplicates - we can't use the main list because of object hashing
 		final String argument = cmdCtx.getArgument(key, String.class);
 		final String[] strArr = argument.split(Pattern.quote(delimiter));
 		final StringReader context = new StringReader(argument);
 		for (String str : strArr) {
 			if (!values.containsKey(str)) {
 				throw new SimpleCommandExceptionType(new LiteralMessage("Item is not allowed in list")).createWithContext(context);
-			} else if (!allowDuplicates && list.contains(values.get(str))) {
+			} else if (!allowDuplicates && listKeys.contains(str)) {
 				throw new SimpleCommandExceptionType(new LiteralMessage("Duplicate arguments are not allowed")).createWithContext(context);
 			} else {
 				list.add(values.get(str));
+				listKeys.add(str);
 			}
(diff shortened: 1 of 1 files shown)


docssrc/src/argument_list.md
```diff
@@ -57,6 +57,20 @@ $$\downarrow$$
 
 $$\downarrow$$
 
+> ### Allowing any value (Optional)
+>
+> If you want to do your own validation you can enable the `allowAnyValue` method to set whether any value is allowed. By default, this is disabled.
+>
+> When `allowAnyValue` is enabled, validation is skipped, and it is expected you will manually perform any validation required:
```
DerEchtePilz, line 64:
```suggestion
> When `allowAnyValue` is set to `true`, list validation is skipped, and any value a player enters will be allowed in the list regardless of the list provided for the argument. Note that duplicate values are still only permitted when `allowDuplicates` is set to `true` too:
```
Maybe this is a bit better?

docssrc/src/argument_list.md
```diff
@@ -57,6 +57,20 @@ $$\downarrow$$
...
+>
+> ```java
+> ListArgumentBuilder.allowAnyValue(true);
+> ```
+>
+> ![List arguments with anyvalue enabled](./images/arguments/listargument_allow_any_values.gif)
```
DerEchtePilz, line 70:
```suggestion
> ![List arguments with allowAnyValue set to true](./images/arguments/listargument_allow_any_values.gif)
```

DerEchtePilz (changes requested):
Just some suggestions I have but nothing that necessarily has to happen if we think the current state is good enough.

docssrc/src/argument_list.md
JorelAli: I don't particularly like the phrase "list validation", we could just be clear as day and remove it. By saying "... is set to `true`, any value a player enters ...", it's so much cleaer.

docssrc/src/argument_list.md
```diff
@@ -57,6 +57,20 @@ $$\downarrow$$
 
 $$\downarrow$$
 
+> ### Allowing any value (Optional)
+>
+> If you want to disable list validation you can enable the `allowAnyValue` method to set whether any value is allowed. By default, this is disabled.
+>
+> When `allowAnyValue` is set to `true`, any value a player enter will be allowed in the list regardless of the list provided for the argument. Note that duplicate values are still only permitted when `allowDuplicates` is set to `true` too
```
DerEchtePilz, line 64:
```suggestion
> When `allowAnyValue` is set to `true`, any value a player enters will be allowed in the list regardless of the list provided for the argument. Note that duplicate values are still only permitted when `allowDuplicates` is set to `true` too
```

JorelAli: Full stop?

DerEchtePilz: No, needs a colon here, I forgot that

DerEchtePilz (approved):
Approving this for now, but I still want to run some more manual tests before merging to make sure it all works correctly.

DerEchtePilz:
Hmm, continuing with the test command I had from #570, when I execute this command
```java
/listtest skip a c e f b d w r t
```

I get the `Duplicate values are not allowed` error message so something still seems to be missing.

Issue #356, opened 2022-10-25
Instructions for testing commands with MockBukkit


JorelAli:
### Description

Instructions for testing the CommandAPI with MockBukkit. Some worked examples, how to set up a project for testing, inclusion of the plugin-test dependency and support for plugins that are shading the CommandAPI are all stuff that would need to be done.

tatanpoker09:
Is there already a plan to incorporate these instructions onto the documentation? I was looking into it myself but I'm having trouble with nms. I am currently trying to copy a little from the plugin-test folder, but a detailed tutorial to avoid having unnecessary bloat would be appreciated

JorelAli:
> Is there already a plan to incorporate these instructions onto the documentation? I was looking into it myself but I'm having trouble with nms. I am currently trying to copy a little from the plugin-test folder, but a detailed tutorial to avoid having unnecessary bloat would be appreciated

Yes! When a suitable framework structure has been made and we're ready to release the public testing setup for CommandAPI commands, we'll add it to the main documentation :)

fhnaumann:
Is there an estimate when this will be available?

willkroboth:
For those following this issue, #575 has been merged to the development branch. Command tests with MockBukkit are now possible using the `commandapi-bukkit-test-toolkit:9.6.2-SNAPSHOT` dependency. Instructions for how to access the development snapshot can be found on the [CommandAPI Discord](https://discord.com/channels/745416925924032523/1003382482944671804/1300605343654481921).

PR #575, opened 2024-07-10
NMS independent test toolkit to allow developers to test their commands with MockBukkit

willkroboth:
Aims to resolve https://github.com/JorelAli/CommandAPI/issues/356.

Currently, when developers try to load the CommandAPI in a MockBukkit environment, it is very hard to get the NMS code accessed by the CommandAPI to cooperate with MockBukkit. There is work on the [`dev/public-test-suite`](https://github.com/JorelAli/CommandAPI/compare/dev/dev...dev/public-test-suite) branch that is trying to make it easier to resolve these issues. As an alternative solution, this PR aims to create a framework for mocking an NMS-independent version of the CommandAPI. Various utilities are also included to make testing commands easier.

These features are accessible using the `commandapi-bukkit-test-toolkit` module. Developers should declare a test scope dependency on this module before their dependency on the plugin or shaded CommandAPI. This overrides CommandAPIVersionHandler when running tests to load MockCommandAPIBukkit instead, which doesn't rely on any NMS classes.

The class CommandAPITestUtilities provides many static methods for asserting particular behavior when running commands. The class CommandAPIHandlerSpy intercepts command registration in the background to enable verifying the arguments input when running a command without having to make any changes to the plugin code that registers commands.

In order to mock Arguments that usually rely upon NMS ArgumentTypes, we would have to implement the parsing logic ourselves. I've included IntegerRangeArgumentType as a proof of concept. Currently, any unimplemented NMS methods throw an UnimplementedMethodException. I don't think we need to worry about specific implementations for all these methods right now. As developers try to use this framework, the methods they actually use can be implemented first.

I've implemented two example projects that showcase how to set up and use the test toolkit, one that shades the CommandAPI and one that depends on the CommandAPI plugin. This also helps verify that each setup works. To more rigorously test the toolkit's features, additional tests have been added inside the `src/test` directory of the `commandapi-bukkit-test-toolkit` module. The result of these tests will appear in the Jacoco coverage report.

I still need to add javadocs and documentation, but I think this is a good start feature-wise. There are obviously many methods that throw an UnimplementedMethodException, but again I think developers can try using this first and show what methods are actually used. Feel free to suggest anything that should be included in an initial snapshot release of the framework though. Also, of course, code review is greatly appreciated!


willkroboth:
Actually, I thought of one important system that should probably be implemented: default ArgumentType suggestions. The IntegerRangeArgument doesn't suggest anything by default, so I didn't think about it at first. There is a parser system in place to make creating new ArgumentType parsers easier, and I can probably add suggestions to that as well. I think I'll add support for the PlayerArgument, since that provides suggestions by default and is commonly used any way.

willkroboth:
Alright, ArgumentType suggestions implemented and docs written. After code review, I'm happy to merge this so developers can try it and figure out which important methods I haven't implemented :P.

DerEchtePilz (commented):
I have absolutely no idea how the `Parser` class works so I just left it alone for now.

Noticed a lot of Spigot API - is there a reason we can't use Paper?
Also when there were selectors, `@n` was missing almost every time. Maybe I missed something while reviewing, but for now I marked those places.

willkroboth:
> Also when there were selectors, `@n` was missing almost every time.

Ah, yeah good catch. I wrote most of this for 1.20.6, and `@n` was added in 1.21 so I didn't include it. I'll definitely update all the dependencies to match 1.21 and add `@n`.
Although, I wonder if this framework should provide some way to account for version-specific behavior? Maybe default to latest, but add something similar to `CommandAPIVersionHandler#usePlatformImplementation` to set a version. Shouldn't be too hard to add that.

> Noticed a lot of Spigot API - is there a reason we can't use Paper?

I was originally thinking that since the CommandAPI is supposed to work on Spigot, the testing framework should work on Spigot, so the Spigot repo makes sure the testing framework doesn't accidentally use Paper-only API. However, I now know that MockBukkit is based on Paper API, so we actually end up having access to Paper API transitively. Technically, we could completely not list a dependency for Spigot/Paper and just transitively inherit it from MockBukkit, which might actually be good because MockBukkit depends on the correct version of the classes being loaded.

So yeah, since MockBukkit uses Paper API this testing framework probably should too?

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/MockCommandAPIPlugin.java
```diff
@@ -0,0 +1,60 @@
...
+	public static MockCommandAPIPlugin load(Consumer<CommandAPIBukkitConfig> configureSettings) {
+		MockCommandAPIPlugin.configureSettings = configureSettings;
+		return MockBukkit.load(MockCommandAPIPlugin.class);
+	}
+
+	@Override
+	public void onLoad() {
+		CommandAPIBukkitConfig config = new CommandAPIBukkitConfig(this);
+
+		if (configureSettings != null) configureSettings.accept(config);
```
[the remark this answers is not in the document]
DerEchtePilz: Oh, if you're doing this change, consider to also do the same in `Parser#parse(StringReader reader)`:
https://github.com/JorelAli/CommandAPI/blob/dev/simple-test-suite/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/arguments/parser/Parser.java#L53-L57

and `ArgumentUtilities#assertCanRead(Function)`:
https://github.com/JorelAli/CommandAPI/blob/dev/simple-test-suite/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/arguments/ArgumentUtilities.java#L74-L78

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/arguments/IntegerRangeArgumentType.java
```diff
@@ -0,0 +1,144 @@
...
+								// Input exact
+								reader -> {
+									int exact = getLow.get();
+									return new IntegerRange(exact, exact);
+								}
+							))
+						)
+				)
+			)
+		);
```
DerEchtePilz, line 134:
```suggestion
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/arguments/IntegerRangeArgumentType.java
```diff
@@ -0,0 +1,144 @@
...
+
+	@Override
+	public IntegerRange parse(StringReader reader) throws CommandSyntaxException {
+		return PARSER.parse(reader);
```
DerEchtePilz, line 138:
```suggestion
		if (!reader.canRead()) {
			throw EMPTY_INPUT.createWithContext(reader);
		}
		int start = reader.getCursor();
		try {
			Result<Integer> maybeLow = Result.withValue(readNumber(reader));
			Result<Integer> maybeHigh;
			if (reader.canRead(2) && (reader.peek() == '.' && reader.peek(1) == '.')) {
				reader.skip();
				reader.skip();
				maybeHigh = Result.withValue(readNumber(reader));
				if (maybeLow.getValue() == null && maybeHigh.getValue() == null) {
					throw EMPTY_INPUT.createWithContext(reader);
				}
			} else {
				maybeHigh = maybeLow;
			}
			if (maybeLow.getValue() == null && maybeHigh.getValue() == null) {
				throw EMPTY_INPUT.createWithContext(reader);
			} else {
				int low = maybeLow.getValue() == null ? Integer.MIN_VALUE : maybeLow.getValue();
				int high = maybeHigh.getValue() == null ? Integer.MAX_VALUE : maybeHigh.getValue();
				if (low > high) {
					throw RANGE_SWAPPED.createWithContext(reader);
				}
				return new IntegerRange(low, high);
			}
		} catch (CommandSyntaxException e) {
			reader.setCursor(start);
			throw new CommandSyntaxException(e.getType(), e.getRawMessage(), e.getInput(), start);
		}
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/arguments/IntegerRangeArgumentType.java
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	public static IntegerRange getRange(CommandContext<MockCommandSource> cmdCtx, String key) {
+		return cmdCtx.getArgument(key, IntegerRange.class);
+	}
```
DerEchtePilz, line 143:
```suggestion
	}

	private Integer readNumber(StringReader reader) throws CommandSyntaxException {
		int cursor = reader.getCursor();

		while (reader.canRead() && isAllowedNumber(reader)) {
			reader.skip();
		}

		String number = reader.getString().substring(cursor, reader.getCursor());
		if (number.isEmpty()) {
			return null;
		} else {
			try {
				return Integer.parseInt(number);
			} catch (NumberFormatException e) {
				throw CommandSyntaxException.BUILT_IN_EXCEPTIONS.readerInvalidInt().createWithContext(reader, number);
			}
		}
	}

	private boolean isAllowedNumber(StringReader reader) {
		char c = reader.peek();
		return (c >= '0' && c <= '9')
			|| (c == '-')
			|| (c == '.')
			&& (!reader.canRead(2) || reader.peek(1) != '.');
	}
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/arguments/parser/Result.java
```diff
@@ -0,0 +1,86 @@
...
+	protected final CommandSyntaxException exception;
+	protected final int suggestionsStart;
+	protected final SuggestionProvider suggestions;
+
+	private Result(T value, CommandSyntaxException exception, int suggestionsStart, SuggestionProvider suggestions) {
+		this.value = value;
+		this.exception = exception;
+		this.suggestionsStart = suggestionsStart;
+		this.suggestions = suggestions;
+	}
```
DerEchtePilz, line 34:
```suggestion
	}

	public T getValue() {
		return value;
	}

	public CommandSyntaxException getException() {
		return exception;
	}
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/test/java/dev/jorel/commandapi/arguments/IntegerRangeArgumentTests.java
```diff
@@ -0,0 +1,159 @@
...
+			player, "test 1.0..",
+			"Invalid integer '1.0' at position 5: test <--[HERE]"
+		);
+		assertCommandFails(
+			player, "test 0..1.0",
+			"Invalid integer '1.0' at position 8: test 0..<--[HERE]"
+		);
+		assertCommandFails(
+			player, "test ..1.0",
+			"Invalid integer '1.0' at position 7: test ..<--[HERE]"
```
DerEchtePilz, line 88:
```suggestion
			"Invalid integer '1.0' at position 5: test <--[HERE]"
		);
		assertCommandFails(
			player, "test ..1.0",
			"Invalid integer '1.0' at position 5: test <--[HERE]"
```

DerEchtePilz (changes requested):
As written on Discord yesterday, I have now written up a complete example without using the `Parser` system as I don't think that system is necessary.

The parsing logic has been adapted from Vanilla code and should basically be a 1:1 copy. This could also be implemented more generally to account for the `FloatRangeArgument` if/when implementing it.
The changes in the test class were results from ingame tests where the position is before the range input.

I think it would be good to do something similar with the other argument types.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/UnimplementedMethodException.java
```diff
@@ -0,0 +1,17 @@
+package dev.jorel.commandapi;
+
+/**
+ * An exception thrown when a method is called that hasn't been implemented yet.
+ */
+// This is the first way I thought to highlight where methods are not implemented
+//  Though all the warnings may be a little unnecessary ¯\_(ツ)_/¯
+@Deprecated(since = "TODO: Implement this method")
```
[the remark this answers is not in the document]
willkroboth: Not all unimplemented methods are located in an obvious `UNIMPLENTED METHODS` section. For example, `EntitySelectorParser` doesn't implement any logic for parsing selector options (the stuff inside `[ ]`), so there is currently a placeholder method that [throws](https://github.com/JorelAli/CommandAPI/blob/d7c798e233516480f779d237cf73abe5da68631d/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/arguments/EntitySelectorParser.java#L109) an `UnimplementedMethodException`. `MockCommandAPIBukkit` has a lot of unimplemented methods, but `MockCommandRegistrationStrategy` also has a few. The warning is _maybe_ a little more helpful in those cases. You get a little, "hey, something is wrong with the implementation of this method," when compiling.

Deprecated is the first thing I thought of that _kinda_ made sense. Deprecated means that using the annotated code is bad should be replaced with a new supported implementation. So, anywhere that `UnimplementedMethodException` is used is bad, and the throw should be replaced with the real implementation. This is definitely non-standard usage of `@Deprecated` though :P. If there's a better annotation that gets the intended message across I'd definitely like to use that.

But yeah, not entirely sure if this makes sense. The warning when compiling and the highlight in an IDE are nice, but the way it's done is certainly strange.

docssrc/src/test_setup.md
```diff
@@ -0,0 +1,89 @@
+# Setting up your project
+
+In the most common simple case, tests can be added directly next to plugin code. Code that goes in your final jar file is located in the `/src/main/` directory, while tests go in `/src/test/`. You can find more information about setting up your project for tests in the [JUnit](https://junit.org/junit5/docs/current/user-guide/#overview-getting-started-example-projects) documentation.
+
+## Dependencies
+
+When you add the dependencies for MockBukkit and `commandapi-bukkit-test-toolkit`, make sure to place them before your main dependencies for the CommandAPI and Spigot/Paper API. This ensures that certain classes that are compatible with the testing environment override the regular classes when running tests.
+
+<div class="multi-pre">
```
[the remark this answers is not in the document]
willkroboth: Added. I think I did it correctly, but I'll leave this unresolved as a reminder until someone checks it :P.

docssrc/src/test_setup.md
```diff
@@ -0,0 +1,90 @@
...
+
+```groovy,Gradle_(build.gradle)
+dependencies {
+    // See https://github.com/MockBukkit/MockBukkit?tab=readme-ov-file#mag-usage for latest version
+    testImplementation 'com.github.seeseemelk:MockBukkit-v1.21:3.128.0'
+
+    testImplementation 'dev.jorel.commandapi-bukkit-test-toolkit:9.6.0-SNAPSHOT'
+
+    // May be the shade dependency and/or mojang-mapped
+    implementation 'dev.jorel:commandapi-bukkit-plugin:9.6.0-SNAPSHOT'
```
DerEchtePilz, line 62: This should be `commandapi-bukkit-core`

docssrc/src/test_setup.md
```diff
@@ -0,0 +1,90 @@
...
+
+```kotlin,Kotlin_Gradle_(build.gradle.kts)
+dependencies {
+    // See https://github.com/MockBukkit/MockBukkit?tab=readme-ov-file#mag-usage for latest version
+    testImplementation('com.github.seeseemelk:MockBukkit-v1.21:3.128.0')
+
+    testImplementation('dev.jorel.commandapi-bukkit-test-toolkit:9.6.0-SNAPSHOT')
+
+    // May be the shade dependency and/or mojang-mapped
+    implementation('dev.jorel:commandapi-bukkit-plugin:9.6.0-SNAPSHOT')
```
DerEchtePilz, line 80: Same here, this should be `commandapi-bukkit-core`

examples/bukkit/automated-tests-shaded/pom.xml
```diff
@@ -0,0 +1,117 @@
...
+			<version>9.6.0-SNAPSHOT</version>
+			<scope>compile</scope>
+		</dependency>
+
+		<!-- Adds the Spigot API to write the plugin -->
+		<!-- If your plugin focuses on Paper, putting that dependency instead works as well -->
+		<dependency>
+			<groupId>org.spigotmc</groupId>
+			<artifactId>spigot-api</artifactId>
+			<version>1.21-R0.1-SNAPSHOT</version>
```
DerEchtePilz, line 52: I guess this could be the 1.21.1 dependency. Not a big deal but for consistency it would be nice.

update.sh
```diff
@@ -21,6 +22,8 @@ sed -i "s/dev\.jorel:commandapi-bukkit-core:$oldVer/dev\.jorel:commandapi-bukkit
 sed -i "s/dev\.jorel:commandapi-annotations:$oldVer/dev\.jorel:commandapi-annotations:$newVer/" docssrc/src/setup_annotations.md
 sed -i "s/dev\.jorel:commandapi-bukkit-kotlin:$oldVer/dev\.jorel:commandapi-bukkit-kotlin:$newVer/" docssrc/src/kotlinintro.md
 sed -i "s/dev\.jorel:commandapi-velocity-kotlin:$oldVer/dev\.jorel:commandapi-velocity-kotlin:$newVer/" docssrc/src/kotlinintro.md
+sed -i "s/dev\.jorel:commandapi-bukkit-plugin:$oldVer/dev\.jorel:commandapi-bukkit-test-toolkit:$newVer/" docssrc/src/test_setup.md
```
DerEchtePilz, line 25: This isn't correct, is it? Shouldn't it be like this:
```suggestion
sed -i "s/dev\.jorel:commandapi-bukkit-test-toolkit:$oldVer/dev\.jorel:commandapi-bukkit-test-toolkit:$newVer/" docssrc/src/test_setup.md
```

DerEchtePilz (changes requested):
Generally good!

In `test_setup.md` you have now added the Gradle dependencies which is good. However, the CommandAPI and Spigot API dependencies are set to `implementation` while they should probably be `compileOnly` which is Gradle's equivalent to Maven's `provided` scope.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/arguments/parser/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+	 * an exception explaining why the input could not be interpreted.
+	 *
+	 * @param reader The {@link StringReader} that holds the input to parse.
+	 * @return The object represented by the given input.
+	 * @throws CommandSyntaxException If the input is malformed.
+	 */
+	default T parse(StringReader reader) throws CommandSyntaxException {
+		Result<T> result = getResult(reader);
+		return result.throwOrReturn();
+	}
```
DerEchtePilz, line 53: This could probably be changed to be one line.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/main/java/dev/jorel/commandapi/arguments/EntitySelectorParser.java
```diff
@@ -0,0 +1,199 @@
...
+			maxResults, includesEntities, order, selfSelector, playerName,
+			entityCheck, entityUUID, type
+		);
+	}
+
+	// Parsing
+	private static ParserLiteral parseSelector(EntitySelectorParser selectorBuilder) {
+		return Parser.read(reader -> {
+			reader.skip(); // skip @
+			if (!reader.canRead()) throw ERROR_MISSING_SELECTOR_TYPE.createWithContext(reader);
```
DerEchtePilz, line 56: Curly braces could probably be added here

DerEchtePilz (approved):
I've left a few points that can be addressed or not, it doesn't really matter.

I also want to say that I really like the form of the `Parser` system. After you pushed that commit I pulled it and looked through it. It was really easy to read the parsers for the different types that are implemented so far. We still got the advantage of handling parsing and suggestions in one object and are able to have a modular system. However, for the most part it's just implementing a lambda and utilizing the `StringReader`.

So, this looks good to me!

willkroboth:
With the 1.21.2/1.21.3 version release coming soon, I'll hold off on merging this (note to self: will have to update CommandAPI dependency versions manually after the release is done). Since this is just a framework with many methods left unimplemented, I think it would be good to release it in a snapshot version first. Developers can try it out and show which methods are important to implement before a full release.

PR #576, opened 2024-07-12
Dev/dev

stumper66:
Added a new option: allowAnyValue


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentBuilder.java
```diff
@@ -129,6 +141,9 @@ public ListArgumentBuilderFinished withStringMapper() {
 		 * @return this list argument builder
 		 */
 		public ListArgumentBuilderFinished withMapper(Function<T, String> mapper) {
+			if (allowAnyValue && !(mapper instanceof StringTooltip)){
```
DerEchtePilz, line 144: This instanceof check is not necessary as it will always evaluate to `true`.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentBuilder.java
```diff
@@ -129,6 +141,9 @@ public ListArgumentBuilderFinished withStringMapper() {
...
+				throw new IllegalArgumentException("allowAnyValue cannot be used with non-string mappers");
```
DerEchtePilz, line 145: I'd say this should be something like:
`allowAnyValue can only be set to true if a String mapper is used.`

However, I do not know if that even is better so maybe we could wait for other opinions too.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
```diff
@@ -137,12 +138,18 @@ public <CommandSourceStack> List<T> parseArgument(CommandContext<CommandSourceSt
...
-			if (!values.containsKey(str)) {
+			if (!allowAnyValue && !values.containsKey(str)) {
 				throw new SimpleCommandExceptionType(new LiteralMessage("Item is not allowed in list")).createWithContext(context);
 			} else if (!allowDuplicates && listKeys.contains(str)) {
 				throw new SimpleCommandExceptionType(new LiteralMessage("Duplicate arguments are not allowed")).createWithContext(context);
 			} else {
-				list.add(values.get(str));
+				if (allowAnyValue) {
+					// should only be used with string mappers so we can ignore the cast warning
+					list.add((T) str);
```
DerEchtePilz, line 148:
```suggestion
					list.add(values.get(str));
```
As we have the map, I think we should use it instead of resorting to generic casts.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
```diff
@@ -137,12 +138,18 @@ public <CommandSourceStack> List<T> parseArgument(CommandContext<CommandSourceSt
...
+				}
+				else{
```
DerEchtePilz, line 150:
```suggestion
				else {
```

docssrc/src/argument_list.md
```diff
@@ -57,6 +57,20 @@ $$\downarrow$$
 
 $$\downarrow$$
 
+> ### Allowing any value (Optional)
```
DerEchtePilz, line 60: In this section I think we should mention that the `allowAnyValue` option only works for Strings.

DerEchtePilz (changes requested):

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
stumper66: @DerEchtePilz when I use the mapper it will give me a list of nulls. The only way I've found that works is to directly cast it to a string.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/arguments/ListArgumentCommon.java
stumper66: @DerEchtePilz when I use the mapper it will give me a list of nulls. The only way I've found that works is to directly cast the string to T.

DerEchtePilz: Well, that sounds like a bug to me that should get fixed. I am currently not available to test anything related to the `ListArgument` but using the mapper should most definitely not result in `null` values.

stumper66: I'm not sure if it's a bug but maybe just a design flaw. The mapper value only works if the input is in the suggestions list.
That part might not be compatible with what we're trying to accomplish so that's why we made allowAnyValue only valid with strings since we need to cast the string input to T.

PR #580, opened 2024-07-21
Made NullCommandSender not throw an exception on Paper

Mrredstone5230:
I basically realized that when trying to use a non-vanilla command on Paper in minecraft functions with CommandAPI (I know, pretty specific), I would get a RybtuneException thrown and the function file wouldn't load
```
[16:50:49 ERROR]: Failed to load function code:load
java.util.concurrent.CompletionException: java.lang.RuntimeException: Failed to wrap CommandSender io.papermc.paper.brigadier.NullCommandSender@dcfd0b2 to a CommandAPI-compatible BukkitCommandSender
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315) ~[?:?]
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320) ~[?:?]
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1770) ~[?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[?:?]
Caused by: java.lang.RuntimeException: Failed to wrap CommandSender io.papermc.paper.brigadier.NullCommandSender@dcfd0b2 to a CommandAPI-compatible BukkitCommandSender
	at CommandAPI-9.5.1.jar/dev.jorel.commandapi.CommandAPIBukkit.wrapCommandSender(CommandAPIBukkit.java:411) ~[CommandAPI-9.5.1.jar:?]
	at CommandAPI-9.5.1.jar/dev.jorel.commandapi.nms.NMS_Common.getCommandSenderFromCommandSource(NMS_Common.java:370) ~[CommandAPI-9.5.1.jar:?]
	at CommandAPI-9.5.1.jar/dev.jorel.commandapi.nms.NMS_Common.getCommandSenderFromCommandSource(NMS_Common.java:91) ~[CommandAPI-9.5.1.jar:?]
	at CommandAPI-9.5.1.jar/dev.jorel.commandapi.CommandAPIHandler.lambda$generatePermissions$1(CommandAPIHandler.java:367) ~[CommandAPI-9.5.1.jar:?]
	at com.mojang.brigadier.tree.CommandNode.canUse(CommandNode.java:81) ~[paper-1.20.6.jar:1.20.6-148-20f5165]
	at com.mojang.brigadier.CommandDispatcher.parseNodes(CommandDispatcher.java:302) ~[paper-1.20.6.jar:?]
	at com.mojang.brigadier.CommandDispatcher.parse(CommandDispatcher.java:292) ~[paper-1.20.6.jar:?]
	at net.minecraft.commands.functions.CommandFunction.parseCommand(CommandFunction.java:103) ~[paper-1.20.6.jar:1.20.6-148-20f5165]
	at net.minecraft.commands.functions.CommandFunction.fromLines(CommandFunction.java:84) ~[paper-1.20.6.jar:1.20.6-148-20f5165]
	at net.minecraft.server.ServerFunctionLibrary.lambda$reload$2(ServerFunctionLibrary.java:91) ~[paper-1.20.6.jar:1.20.6-148-20f5165]
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768) ~[?:?]
	... 3 more
```

After doing some digging, I found where that Exception was thrown, and realized what was happening. Looking at nms code, Minecraft loads functions with this NullCommandSender (but when they are called they do use the correct sender). This PR includes a simple test that avoids the exception by throwing null (which is only used by the CommandAPIHandler.permissionCheck(..) method, which already supports a null sender.) This fixed my issue.

(I couldn't replicate my issue on Spigot, so this fix being specific to paper is probably fine)

Tested with PaperMC 1.20.6 (latest at the time) and CommandAPI 9.5.1.


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
```diff
@@ -409,6 +409,13 @@ public BukkitCommandSender<? extends CommandSender> wrapCommandSender(CommandSen
 				// We literally cannot type this at compile-time, so let's use a placeholder CommandSender instance
 				return new BukkitFeedbackForwardingCommandSender<CommandSender>(FeedbackForwardingSender.cast(sender));
 			}
+
+			final Class<? extends CommandSender> NullCommandSender = paper.getNullCommandSender();
+			if (NullCommandSender.isInstance(sender)) {
```
willkroboth, line 414: One small concern: I spot an NPE here. If we're running a version of Paper that does not include the NullCommandSender class (anything before 1.20.6), then `isPaperPresent` will be true, but `nullCommandSender` will be null. So, `NullCommandSender#isInstance` will NPE.

This _would_ only happen if a CommandSender implementation that does not match any of the previous cases was passed into here. That probably wouldn't happen, but if it did, we want the exception to be the `Failed to wrap CommandSender` exception thrown below instead of a general NPE.

So maybe add a `null` check here (and a note to the javadoc)? It's a little weird when Paper-specific features are _also_ version dependent, but I think this is fine.

```suggestion
			final Class<? extends CommandSender> NullCommandSender = paper.getNullCommandSender();
			if (NullCommandSender != null && NullCommandSender.isInstance(sender)) {
```

willkroboth (commented):
Looking at the Paper source, yep, this is a new Paper thing. Their Brigadier command API PR (https://github.com/PaperMC/Paper/pull/8235) [created](https://github.com/PaperMC/Paper/blob/aa36ae65764f652a16517a994e8d9cf4a89136c8/patches/server/0972-Brigadier-based-command-API.patch#L61-L217) the `NullCommandSender` class and changed the Null CommandSource Minecraft uses for parsing functions to [use that sender](https://github.com/PaperMC/Paper/blob/aa36ae65764f652a16517a994e8d9cf4a89136c8/patches/server/0972-Brigadier-based-command-API.patch#L2014). So, this change should only be necessary on paper-1.20.6-65 and later. I confirmed that CommandAPI 9.5.1 throws the RuntimeException on Paper 1.20.6 and 1.21, but not on paper-1.20.6-64. This PR fixes the issue, so the function loads and runs CommandAPI commands successfully.

Spigot 1.20.6 and 1.21 can successfully load a datapack that contains CommandAPI commands. I was a bit confused about how this was possible, given that on Spigot, instead of returning a `NullCommandSender`, the same method throws a `UnsupportedOperationException`, which I expected to bubble up. What I didn't realize is that the CommandAPI actually catches this exception and returns null: 

https://github.com/JorelAli/CommandAPI/blob/d130bdee6647cb046fd326058b96cedfa56de59d/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-nms-common/src/main/java/dev/jorel/commandapi/nms/NMS_Common.java#L367-L374

Looks like we already had some logic I didn't know about to deal with the special command source used when parsing functions, but Paper did thier change and now we need extra, special logic. I never thought about what that bit of code was supposed to do before :P. I was a bit skeptical when I saw your code return null from `wrapCommandSender`, but that seems to be the correct choice given that is what happens on Spigot.

So, uh, long story short, good changes :+1:. Just the one comment.

Mrredstone5230:
Great! Thanks so much for the feedback!

willkroboth (approved):

PR #594, opened 2024-08-12
Be more lenient with version checks when updating the minor version of the game

DerEchtePilz:
Currently, when updating to *any* new Minecraft version that is not currently supported, we throw an `UnsupportedVersionException`.
This is not that great since even minor updates require an update to not throw that exception.

This pull request enables, when setting the config option, an update to minor versions of the game:
- An update from 1.21.1 to 1.21.2 would be allowed
- An update from 1.21.2 to 1.22 would still result in an `UnsupportedVersionException` since presumably we still want to double check compatibility for major versions.

**So why the new config option when we have the option of just loading the latest NMS version?**
Well, the latest NMS version completely breaks support for older versions, this only does something when a server version is used that the main part of `CommandAPIVersionHandler#getPlatform()` does not handle currently.

**Further considerations:**
Currently, this does not include a new config option for the plugin version but I think this is something that could and should be done before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,7 +414,7 @@ This is the current roadmap for the CommandAPI (as of 30th April 2024):
             <td valign="top">???</td>
             <td valign="top">
                 <ul>
-                    <li>https://github.com/JorelAli/CommandAPI/pull/594 Adds a config option to allow the CommandAPI to be more lenient when updating to a new minor version (e. g. from 1.21 to 1.21.1)</li>
+                    <li>https://github.com/JorelAli/CommandAPI/pull/594 Adds a config option to allow the CommandAPI to be more lenient when updating to a new minor version (e.g. from 1.21 to 1.21.1)</li>
                 </ul>
             </td>
         </tr>
--- a/README.md
+++ b/README.md
@@ -410,6 +410,15 @@ This is the current roadmap for the CommandAPI (as of 30th April 2024):
     </thead>
     <tbody>
         <tr>
+            <td valign="top"><b>9.6.0</b></td>
+            <td valign="top">???</td>
+            <td valign="top">
+                <ul>
+                    <li>https://github.com/JorelAli/CommandAPI/pull/594 Adds a config option to allow the CommandAPI to be more lenient when updating to a new minor version (e. g. from 1.21 to 1.21.1)</li>
+                </ul>
+            </td>
+        </tr>
+        <tr>
             <td valign="top"><b>9.5.3</b></td>
             <td valign="top">August 2024</td>
             <td valign="top">
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
@@ -112,6 +112,11 @@ public class CommandAPI {
 			CommandAPIPlatform<?, ?, ?> platform = CommandAPIVersionHandler.getPlatform();
 			new CommandAPIHandler<>(platform);
 
+			if (CommandAPI.getConfiguration().shouldUseLatestNMSVersion() || CommandAPI.getConfiguration().shouldBeLenientForMinorVersions()) {
+				CommandAPI.logWarning("Loading the CommandAPI with a potentially incompatible NMS implementation.");
+				CommandAPI.logWarning("While you may find success with this, further updates might be necessary to fully support the version you are using.");
+			}
+
 			// Log platform load
 			final String platformClassHierarchy;
 			{
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
@@ -109,13 +109,10 @@ public class CommandAPI {
 			CommandAPI.config = new InternalConfig(config);
 
 			// Initialize handlers
-			CommandAPIPlatform<?, ?, ?> platform = CommandAPIVersionHandler.getPlatform();
+			LoadContext loadContext = CommandAPIVersionHandler.getPlatform();
+			CommandAPIPlatform<?, ?, ?> platform = loadContext.platform();
 			new CommandAPIHandler<>(platform);
-
-			if (CommandAPI.getConfiguration().shouldUseLatestNMSVersion() || CommandAPI.getConfiguration().shouldBeLenientForMinorVersions()) {
-				CommandAPI.logWarning("Loading the CommandAPI with a potentially incompatible NMS implementation.");
-				CommandAPI.logWarning("While you may find success with this, further updates might be necessary to fully support the version you are using.");
-			}
+			loadContext.context().run();
 
 			// Log platform load
 			final String platformClassHierarchy;
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
@@ -39,7 +39,7 @@ extends CommandAPIConfig<Impl>
 	boolean verboseOutput = false;
 	boolean silentLogs = false;
 	boolean useLatestNMSVersion = false;
-	boolean lenientForMinorVersions = false;
+	boolean beLenientForMinorVersions = false;
 	String missingExecutorImplementationMessage = "This command has no implementations for %s";
 
 	File dispatcherFile = null;
@@ -93,7 +93,7 @@ extends CommandAPIConfig<Impl>
 
 	/**
 	 * Sets whether the CommandAPI should load a (potentially unsupported) NMS version
-	 * when updating to a minor release of Minecraft, for example this setting would
+	 * when updating to a minor release of Minecraft. For example, this setting would
 	 * allow running a minor version of the game before the CommandAPI makes a proper
 	 * release for it. Unlike {@link #useLatestNMSVersion(boolean)}, this setting does
 	 * not blindly load the latest NMS version and as a result break cross version compatibility
@@ -102,8 +102,8 @@ extends CommandAPIConfig<Impl>
 	 * @param value whether the CommandAPI should assume that minor Minecraft releases do not cause incompatibilities
 	 * @return this CommandAPIConfig
 	 */
-	public Impl lenientForMinorVersions(boolean value) {
-		this.lenientForMinorVersions = value;
+	public Impl beLenientForMinorVersions(boolean value) {
+		this.beLenientForMinorVersions = value;
 		return instance();
 	}
 
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
@@ -96,8 +96,7 @@ extends CommandAPIConfig<Impl>
 	 * when updating to a minor release of Minecraft. As an example, this setting can allow
 	 * updating to 1.21.2 from 1.21.1 but doesn't allow updating to 1.22 from 1.21.2.
 	 * Unlike {@link #useLatestNMSVersion(boolean)}, this setting does
-	 * not blindly load the latest NMS version and as a result breaks cross version compatibility
-	 * but will actually prefer loading the correct NMS implementation.
+	 * not blindly load the latest NMS version, but will prefer loading the correct NMS implementation when available.
 	 *
 	 * @param value whether the CommandAPI should assume that minor Minecraft releases do not cause incompatibilities
 	 * @return this CommandAPIConfig
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
@@ -39,6 +39,7 @@ extends CommandAPIConfig<Impl>
 	boolean verboseOutput = false;
 	boolean silentLogs = false;
 	boolean useLatestNMSVersion = false;
+	boolean lenientForMinorVersions = false;
 	String missingExecutorImplementationMessage = "This command has no implementations for %s";
 
 	File dispatcherFile = null;
@@ -87,6 +88,22 @@ extends CommandAPIConfig<Impl>
 	 */
 	public Impl useLatestNMSVersion(boolean value) {
 		this.useLatestNMSVersion = value;
+		return instance();
+	}
+
+	/**
+	 * Sets whether the CommandAPI should load a (potentially unsupported) NMS version
+	 * when updating to a minor release of Minecraft, for example this setting would
+	 * allow running a minor version of the game before the CommandAPI makes a proper
+	 * release for it. Unlike {@link #useLatestNMSVersion(boolean)}, this setting does
+	 * not blindly load the latest NMS version and as a result break cross version compatibility
+	 * but will actually prefer loading the correct NMS implementation.
+	 *
+	 * @param value whether the CommandAPI should assume that minor Minecraft releases do not cause incompatibilities
+	 * @return this CommandAPIConfig
+	 */
+	public Impl lenientForMinorVersions(boolean value) {
+		this.lenientForMinorVersions = value;
 		return instance();
 	}
 
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
@@ -93,10 +93,10 @@ extends CommandAPIConfig<Impl>
 
 	/**
 	 * Sets whether the CommandAPI should load a (potentially unsupported) NMS version
-	 * when updating to a minor release of Minecraft. For example, this setting would
-	 * allow running a minor version of the game before the CommandAPI makes a proper
-	 * release for it. Unlike {@link #useLatestNMSVersion(boolean)}, this setting does
-	 * not blindly load the latest NMS version and as a result break cross version compatibility
+	 * when updating to a minor release of Minecraft. As an example, this setting can allow
+	 * updating to 1.21.2 from 1.21.1 but doesn't allow updating to 1.22 from 1.21.2.
+	 * Unlike {@link #useLatestNMSVersion(boolean)}, this setting does
+	 * not blindly load the latest NMS version and as a result breaks cross version compatibility
 	 * but will actually prefer loading the correct NMS implementation.
 	 *
 	 * @param value whether the CommandAPI should assume that minor Minecraft releases do not cause incompatibilities
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/InternalConfig.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/InternalConfig.java
@@ -40,7 +40,7 @@ public class InternalConfig {
 	// Whether we should use the latest NMS version (which may not be compatible)
 	private final boolean useLatestNMSVersion;
 
-	private final boolean beMoreLenientForMinorVerions;
+	private final boolean beLenientForMinorVersions;
 
 	// The message to display when an executor implementation is missing
 	private final String messageMissingExecutorImplementation;
@@ -67,7 +67,7 @@ public class InternalConfig {
 		this.verboseOutput = config.verboseOutput;
 		this.silentLogs = config.silentLogs;
 		this.useLatestNMSVersion = config.useLatestNMSVersion;
-		this.beMoreLenientForMinorVerions = config.lenientForMinorVersions;
+		this.beLenientForMinorVersions = config.beLenientForMinorVersions;
 		this.messageMissingExecutorImplementation = config.missingExecutorImplementationMessage;
 		this.dispatcherFile = config.dispatcherFile;
 		this.skipSenderProxy = config.skipSenderProxy;
@@ -100,8 +100,8 @@ public class InternalConfig {
 	/**
 	 * @return Whether the CommandAPI should assume that minor versions of officially unsupported versions do not cause incompatibilities
 	 */
-	public boolean shouldBeMoreLenientForMinorVersions() {
-		return this.beMoreLenientForMinorVerions;
+	public boolean shouldBeLenientForMinorVersions() {
+		return this.beLenientForMinorVersions;
 	}
 
 	/**
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/InternalConfig.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/InternalConfig.java
@@ -40,6 +40,8 @@ public class InternalConfig {
 	// Whether we should use the latest NMS version (which may not be compatible)
 	private final boolean useLatestNMSVersion;
 
+	private final boolean beMoreLenientForMinorVerions;
+
 	// The message to display when an executor implementation is missing
 	private final String messageMissingExecutorImplementation;
 
@@ -65,6 +67,7 @@ public class InternalConfig {
 		this.verboseOutput = config.verboseOutput;
 		this.silentLogs = config.silentLogs;
 		this.useLatestNMSVersion = config.useLatestNMSVersion;
+		this.beMoreLenientForMinorVerions = config.lenientForMinorVersions;
 		this.messageMissingExecutorImplementation = config.missingExecutorImplementationMessage;
 		this.dispatcherFile = config.dispatcherFile;
 		this.skipSenderProxy = config.skipSenderProxy;
@@ -92,6 +95,13 @@ public class InternalConfig {
 	 */
 	public boolean shouldUseLatestNMSVersion() {
 		return this.useLatestNMSVersion;
+	}
+
+	/**
+	 * @return Whether the CommandAPI should assume that minor versions of officially unsupported versions do not cause incompatibilities
+	 */
+	public boolean shouldBeMoreLenientForMinorVersions() {
+		return this.beMoreLenientForMinorVerions;
 	}
 
 	/**
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin-mojang-mapped/src/main/java/dev/jorel/commandapi/CommandAPIMain.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin-mojang-mapped/src/main/java/dev/jorel/commandapi/CommandAPIMain.java
@@ -51,7 +51,8 @@ public class CommandAPIMain extends JavaPlugin {
 			.missingExecutorImplementationMessage(fileConfig.getString("messages.missing-executor-implementation"))
 			.dispatcherFile(fileConfig.getBoolean("create-dispatcher-json") ? new File(getDataFolder(), "command_registration.json") : null)
 			.shouldHookPaperReload(fileConfig.getBoolean("hook-paper-reload"))
-			.skipReloadDatapacks(fileConfig.getBoolean("skip-initial-datapack-reload"));
+			.skipReloadDatapacks(fileConfig.getBoolean("skip-initial-datapack-reload"))
+			.lenientForMinorVersions(fileConfig.getBoolean("lenient-for-minor-versions"));
 
 		for (String pluginName : fileConfig.getStringList("skip-sender-proxy")) {
 			if (Bukkit.getPluginManager().getPlugin(pluginName) != null) {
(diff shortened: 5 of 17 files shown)


README.md
```diff
@@ -409,6 +409,15 @@ This is the current roadmap for the CommandAPI (as of 30th April 2024):
         </tr>
     </thead>
     <tbody>
+        <tr>
+            <td valign="top"><b>9.6.0</b></td>
+            <td valign="top">???</td>
+            <td valign="top">
+                <ul>
+                    <li>https://github.com/JorelAli/CommandAPI/pull/594 Adds a config option to allow the CommandAPI to be more lenient when updating to a new minor version (e. g. from 1.21 to 1.21.1)</li>
```
Timongcraft, line 417: *Small nitpick*: it should be written as “(e.g. from 1.21 to 1.21.1)” instead of “(e. g. from 1.21 to 1.21.1).”

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -68,9 +70,54 @@ public interface CommandAPIVersionHandler {
...
+			}
 		}
 	}
 
+	class Version {
+
+		private final String version;
+
+		// The minor version is something like 1.x
+		private final int minor;
```
Timongcraft, line 91: Shouldn't this be named major?

DerEchtePilz: Yeah, probably should revise all the names in the PR again. Threw it together fairly quickly and first had the Minecraft version treated as major.minor.patch and just deleted the major and patch fields.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -49,11 +49,13 @@ public interface CommandAPIVersionHandler {
 	 * @return an instance of NMS which can run on the specified Minecraft version
 	 */
 	static CommandAPIPlatform<?, ?, ?> getPlatform() {
+		CommandAPIPlatform<?, ?, ?> latestNMS = new NMS_1_21_R1();
```
willkroboth, line 52: Constructing `latestNMS` here means it always attempts to load when this method is called. This seems to cause an exception on any version other than 1.21/1.21.1 (tested on Paper 1.19.4 and 1.20).

For example, simply adding the CommandAPI plugin to the server and running on Paper 1.20 gives this error:

```
[17:07:23 ERROR]: [CommandAPI] Error initializing plugin 'CommandAPI-9.6.0-SNAPSHOT_12_Aug_2024_(08-42-37PM).jar' in folder 'plugins' (Is it up to date?)
java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    dev/jorel/commandapi/nms/NMS_1_21_R1.<clinit>()V @38: invokestatic
  Reason:
    Type 'net/minecraft/core/IRegistryCustom$Dimension' (current frame, stack[0]) is not assignable to 'net/minecraft/core/HolderLookup$a'
  Current Frame:
    bci: @38
    flags: { }
    locals: { 'org/bukkit/craftbukkit/v1_21_R1/CraftServer', 'org/bukkit/Server' }
    stack: { 'net/minecraft/core/IRegistryCustom$Dimension', 'net/minecraft/world/flag/FeatureFlagSet' }
  Bytecode:
    0000000: b804 c04c 2bc1 04c2 9900 272b c004 c24b
    0000010: 2ab6 0703 b608 752a b607 03b6 0876 b906
    0000020: 8401 00b6 07a1 b808 7bb3 007f a700 0701
    0000030: b300 7f12 e813 087d 1308 7d12 f0b8 0881
    0000040: b300 e113 0344 1308 8313 0885 b808 8bb3
    0000050: 034a 1308 8d13 088e 1308 8fb8 088b b306
    0000060: 6c12 4313 0890 b608 9457 043b a700 064c
    0000070: 033b 1ab3 0711 b1                      
  Exception Handler Table:
    bci [97, 108] => handler: 111
    bci [97, 108] => handler: 111
  Stackmap Table:
    same_frame(@47)
    same_frame(@51)
    same_locals_1_stack_item_frame(@111,Object[#386])
    append_frame(@114,Integer)

        at dev.jorel.commandapi.CommandAPIVersionHandler.getPlatform(CommandAPIVersionHandler.java:52) ~[CommandAPI-9.6.0-SNAPSHOT_12_Aug_2024_(08-42-37PM).jar:?]
        at dev.jorel.commandapi.CommandAPI.onLoad(CommandAPI.java:112) ~[CommandAPI-9.6.0-SNAPSHOT_12_Aug_2024_(08-42-37PM).jar:?]
        at dev.jorel.commandapi.CommandAPIMain.onLoad(CommandAPIMain.java:68) ~[CommandAPI-9.6.0-SNAPSHOT_12_Aug_2024_(08-42-37PM).jar:?]
        at io.papermc.paper.plugin.storage.ServerPluginProviderStorage.processProvided(ServerPluginProviderStorage.java:59) ~[paper-1.20.jar:git-Paper-17]
        at io.papermc.paper.plugin.storage.ServerPluginProviderStorage.processProvided(ServerPluginProviderStorage.java:18) ~[paper-1.20.jar:git-Paper-17]
        at io.papermc.paper.plugin.storage.SimpleProviderStorage.enter(SimpleProviderStorage.java:40) ~[paper-1.20.jar:git-Paper-17]
        at io.papermc.paper.plugin.entrypoint.LaunchEntryPointHandler.enter(LaunchEntryPointHandler.java:36) ~[paper-1.20.jar:git-Paper-17]
        at org.bukkit.craftbukkit.v1_20_R1.CraftServer.loadPlugins(CraftServer.java:430) ~[paper-1.20.jar:git-Paper-17]
        at net.minecraft.server.dedicated.DedicatedServer.initServer(DedicatedServer.java:273) ~[paper-1.20.jar:git-Paper-17]
        at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1101) ~[paper-1.20.jar:git-Paper-17]
        at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:318) ~[paper-1.20.jar:git-Paper-17]
        at java.lang.Thread.run(Thread.java:833) ~[?:?]
```

This prevents the CommandAPI from loading. Perhaps instead of constructing the nms instance to check `NMS#compatibleVersions`, this code could just check if the version matches a hardcoded current latest minor version. This would need to be updated manually, but that seems similar to needing to update the result given by `shouldUseLatestNMSVersion`.

```java
if (shouldUseLatestNMSVersion) {
    return new NMS_1_21_R1();
}

switch (version) {
	case "1.16.5" -> new NMS_1_16_R3();
	// ...
	case "1.21", "1.21.1" -> new NMS_1_21_R1();
    default -> {
        if (shouldBeMoreLenientForMinorVersions && version.split("\\.")[1].equals("21")) {
            yield new NMS_1_21_R1();
        } else {
            throw UnsupportedVersionException(version);
        }
    }
}
```

Just need to avoid constructing `NMS_1_21_R1`  when the switch statement will match a `version` String.

commandapi-core/src/main/java/dev/jorel/commandapi/InternalConfig.java
```diff
@@ -65,6 +67,7 @@ public InternalConfig(CommandAPIConfig<?> config) {
 		this.verboseOutput = config.verboseOutput;
 		this.silentLogs = config.silentLogs;
 		this.useLatestNMSVersion = config.useLatestNMSVersion;
+		this.beMoreLenientForMinorVerions = config.lenientForMinorVersions;
```
willkroboth, line 70: Small note: it's a bit strange these fields have different names, which isn't the case for other config options.

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
```diff
@@ -90,6 +91,22 @@ public Impl useLatestNMSVersion(boolean value) {
 		return instance();
 	}
 
+	/**
+	 * Sets whether the CommandAPI should load a (potentially unsupported) NMS version
+	 * when updating to a minor release of Minecraft, for example this setting would
```
willkroboth, line 96: Should this be a new sentence?

```suggestion
	 * when updating to a minor release of Minecraft. For example, this setting would
```

Also, I like the example of `this setting can allow updating to 1.21.2 from 1.21.1 but doesn't allow updating to 1.22 from 1.21.2` from the docs. I think it makes the meaning of "minor release" very clear given that Minecraft doesn't quite follow the common SemVer logic of `major.minor.patch` (1.21 to 1.22 might seem like a change in the minor version, and hence be confused as a minor update). Perhaps that could be incorporated in these javadocs?

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin-mojang-mapped/src/main/resources/config.yml
```diff
@@ -41,6 +41,13 @@ create-dispatcher-json: false
 # implementation is actually compatible with the current Minecraft version.
 use-latest-nms-version: false
 
+# Be lenient with version checks when loading for new minor Minecraft versions (default: false)
+# If "true", the CommandAPI loads NMS implementations for (potentially unsupported) Minecraft versions.
+# For example, this setting may allow updating from 1.21.1 to 1.21.2 as only the minor version is changing
+# but will not allow an update from 1.21.2 to 1.22
+# Keep in mind that implementations my vary and actually updating the CommandAPI might be necessary.
```
willkroboth, line 48:
```suggestion
# but will not allow an update from 1.21.2 to 1.22.
# Keep in mind that implementations may vary and actually updating the CommandAPI might be necessary.
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin/src/main/resources/config.yml
```diff
@@ -41,6 +41,13 @@ create-dispatcher-json: false
 # implementation is actually compatible with the current Minecraft version.
 use-latest-nms-version: false
 
+# Be lenient with version checks when loading for new minor Minecraft versions (default: false)
+# If "true", the CommandAPI loads NMS implementations for (potentially unsupported) Minecraft versions.
+# For example, this setting may allow updating from 1.21.1 to 1.21.2 as only the minor version is changing
+# but will not allow an update from 1.21.2 to 1.22
+# Keep in mind that implementations my vary and actually updating the CommandAPI might be necessary.
```
willkroboth, line 48: Same note as `plugin-mojang-mapped`.

```suggestion
# but will not allow an update from 1.21.2 to 1.22.
# Keep in mind that implementations may vary and actually updating the CommandAPI might be necessary.
```

(_Sidenote outside of scope, but I wonder if there's an easy way for spigot and mojang-mapped plugin to share the config file resource, since they should be the same_)

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -68,9 +70,54 @@ public interface CommandAPIVersionHandler {
 				case "1.20.3", "1.20.4" -> new NMS_1_20_R3();
 				case "1.20.5", "1.20.6" -> new NMS_1_20_R4();
 				case "1.21", "1.21.1" -> new NMS_1_21_R1();
-				default -> throw new UnsupportedVersionException(version);
+				default -> null;
```
willkroboth, line 73: The `null` here and the intermediate `platform` variable could be avoided using the [`yield`](https://www.baeldung.com/java-yield-switch) keyword:

```java
return switch (version) {
	case "1.16.5" -> new NMS_1_16_R3();
	// ...
	case "1.21", "1.21.1" -> new NMS_1_21_R1();
	default -> {
		if (CommandAPI.getConfiguration().shouldBeMoreLenientForMinorVersions()) {
			yield minecraftVersion.matchesPatch((NMS<?>) latestNMS);
		}
		throw new UnsupportedVersionException(version);
	}
};
```

(_Dunno why GitHub doesn't render it red like the other keywords, it totally exists :P_)

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -68,9 +70,54 @@ public interface CommandAPIVersionHandler {
...
 			};
+			if (platform != null) {
+				return platform;
+			} else {
+				if (CommandAPI.getConfiguration().shouldBeMoreLenientForMinorVersions()) {
+					return minecraftVersion.matchesPatch((NMS<?>) latestNMS);
```
willkroboth, line 79: The cast to `NMS<?>` here could be avoided by declaring `latestNMS` with type `CommandAPIBukkit<?>` rather than `CommandAPIPlatform<?, ?, ?>`, since `CommandAPIBukkit<?>` implements both of those.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -49,11 +49,13 @@ public interface CommandAPIVersionHandler {
 	 * @return an instance of NMS which can run on the specified Minecraft version
 	 */
 	static CommandAPIPlatform<?, ?, ?> getPlatform() {
+		CommandAPIPlatform<?, ?, ?> latestNMS = new NMS_1_21_R1();
 		if (CommandAPI.getConfiguration().shouldUseLatestNMSVersion()) {
-			return new NMS_1_21_R1();
+			return latestNMS;
```
willkroboth, line 54: This applies to `shouldUseLatestNMSVersion` and `shouldBeMoreLenientForMinorVersions`, but maybe there could be a warning message logged to make it clear when we're loading a potentially incompatible NMS version.

willkroboth (changes requested):

README.md
JorelAli: I'd rather we use "for example" instead of "e.g." in the documentation

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin/src/main/resources/config.yml
JorelAli: > (_Sidenote outside of scope, but I wonder if there's an easy way for spigot and mojang-mapped plugin to share the config file resource, since they should be the same_)

All we need is _yet another Maven module_ that they both inherit from 🫠

willkroboth: I suppose one could perhaps inherit from the other. I think the only difference is the nms dependency, so as long as those get filtered correctly ¯\\\_(ツ)_/¯. That could also avoid duplicating CommandAPIMain and AdvancedConvertor.

Might also be possible to tell Maven to pull from the same shared resource folder.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
XHawk87: They refuse to yield to change

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
```diff
@@ -112,6 +112,11 @@ public static void onLoad(CommandAPIConfig<?> config) {
 			CommandAPIPlatform<?, ?, ?> platform = CommandAPIVersionHandler.getPlatform();
 			new CommandAPIHandler<>(platform);
 
+			if (CommandAPI.getConfiguration().shouldUseLatestNMSVersion() || CommandAPI.getConfiguration().shouldBeLenientForMinorVersions()) {
+				CommandAPI.logWarning("Loading the CommandAPI with a potentially incompatible NMS implementation.");
+				CommandAPI.logWarning("While you may find success with this, further updates might be necessary to fully support the version you are using.");
+			}
+
```
willkroboth, line 119: This may mistakenly log the warning when `shouldUseLatestNMSVersion = false`, `shouldBeMoreLenientForMinorVersions = true`, and the plugin is being loaded on a supported version. For example, if I set the config `lenient-for-minor-versions: true` and run the plugin on Paper 1.20, I get this log:

```
[10:04:46 INFO]: [CommandAPI] Loading server plugin CommandAPI v9.6.0-SNAPSHOT
[10:04:46 WARN]: [CommandAPI] Loading the CommandAPI with a potentially incompatible NMS implementation.
[10:04:46 WARN]: [CommandAPI] While you may find success with this, further updates might be necessary to fully support the version you are using.
[10:04:46 INFO]: [CommandAPI] Loaded platform NMS_1_20_R1 > NMS_CommonWithFunctions > NMS_Common > CommandAPIBukkit
```

This seems misleading, as `NMS_1_20_R1` _is_ compatible with Paper 1.20 (it is not potentially incompatible).

Additionally, I'm not sure if it makes sense to put this warning in the `CommandAPI` class, since this code will also run on Velocity. Velocity does not do anything if these options are enabled (for example when shading, since `CommandAPIConfig` provides access to these methods on Velocity), so it will always load a compatible platform.

Perhaps these warning messages could be placed in `commandapi-bukit-vh`

```java
if (shouldUseLatestNMSVersion) {
    warn("Forcefully loading the latest NMS version. This is not guaranteed to work.")
    return NMS_1_21();
}

return switch(version) {
    case "1.16.5" -> NMS_1_16_5();
    // ...
    default -> {
        if (shouldBeMoreLeinientForMinorVersions()) {
            warn("Attempting to load a minor NMS version. This is not guaranteed to work.")
            return NMS_1_21();
        } else {
            throw UnsupportedVersion();
        }
    }
}
```

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
```diff
@@ -90,6 +91,22 @@ public Impl useLatestNMSVersion(boolean value) {
 		return instance();
 	}
 
+	/**
+	 * Sets whether the CommandAPI should load a (potentially unsupported) NMS version
+	 * when updating to a minor release of Minecraft. As an example, this setting can allow
+	 * updating to 1.21.2 from 1.21.1 but doesn't allow updating to 1.22 from 1.21.2.
+	 * Unlike {@link #useLatestNMSVersion(boolean)}, this setting does
+	 * not blindly load the latest NMS version and as a result breaks cross version compatibility
+	 * but will actually prefer loading the correct NMS implementation.
```
willkroboth, line 100: This wording might be clearer?

```suggestion
	 * not blindly load the latest NMS version, but will prefer loading the correct NMS implementation when available.
```

willkroboth (changes requested):

commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPI.java
DerEchtePilz: Well, I originally put this in the version handler but apparently the logger needs a platform to work so I put it after the initialisation of the CommandAPIHandler instance.

What I could do is adjust the return type for the getPlatform method a bit to hold more context, such as the latest supported version and of course the CommandAPIPlatform instance?

willkroboth: Ohhhh, yea, hmmm, that's tricky...

Maybe the extra context could be a `Consumer` to call back like so?

```java
record LoadInfo(CommandAPIPlatform<?, ?, ?> platform, Runnable postLoadCallback);

static LoadInfo getPlatform() {
    if (shouldUseLatestNMSVersion) {
        return new LoadInfo(new NMS_1_21_R1(), () -> {
            CommandAPI.logWarning("Forcefully loading the latest NMS version. This is not guaranteed to work.");
        });
    }

    CommandAPIPlatform<?, ?, ?> platform = switch (version) {
        case "1.16.5" -> new NMS_1_16_R3();
        // ...
        case "1.21", "1.21.1" -> new NMS_1_21_R1();
        default -> null;
    };

    if (platform != null) {
        // Normal load, nothing to do
        return new LoadInfo(platform, () -> {});
    }

    if (shouldBeLenientForMinorVersions) {
        return new LoadInfo(new NMS_1_21_R1(), () -> {
            CommandAPI.logWarning("Attempting to load a minor NMS version. This is not guaranteed to work.");
        });
    } else {
        throw new UnsupportedVersionException(version);
    }
}
```

DerEchtePilz: Yeah, that's a good idea.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin-mojang-mapped/src/main/resources/config.yml
```diff
@@ -41,6 +41,13 @@ create-dispatcher-json: false
 # implementation is actually compatible with the current Minecraft version.
 use-latest-nms-version: false
 
+# Be lenient with version checks when loading for new minor Minecraft versions (default: false)
+# If "true", the CommandAPI loads NMS implementations for (potentially unsupported) Minecraft versions.
+# For example, this setting may allow updating from 1.21.1 to 1.21.2 as only the minor version is changing
+# but will not allow an update from 1.21.2 to 1.22.
+# Keep in mind that implementations may vary and actually updating the CommandAPI might be necessary.
+lenient-for-minor-versions: false
```
willkroboth, line 49: Should this be called `be-lenient-for-minor-versions`, like it is in the code, similar to `use-latest-nms-version`?

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -68,8 +72,23 @@ public interface CommandAPIVersionHandler {
 				case "1.20.3", "1.20.4" -> new NMS_1_20_R3();
 				case "1.20.5", "1.20.6" -> new NMS_1_20_R4();
 				case "1.21", "1.21.1" -> new NMS_1_21_R1();
-				default -> throw new UnsupportedVersionException(version);
+				default -> null;
 			};
+			if (platform != null) {
+				return new LoadContext(platform, () -> {});
+			}
```
willkroboth, line 79: `LoadContext` could have an additional constructor for the case where nothing extra needs to run:

```java
public record LoadContext(CommandAPIPlatform<?, ?, ?> platform, Runnable context) {
	public LoadContext(CommandAPIPlatform<?, ?, ?> platform) {
		this(platform, () -> {});
	}
}
```

That might make situations like this where no extra logic needs to run clearer.

```suggestion
			if (platform != null) {
				return new LoadContext(platform);
			}
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -48,12 +48,16 @@ public interface CommandAPIVersionHandler {
...
 	 * @return an instance of NMS which can run on the specified Minecraft version
 	 */
-	static CommandAPIPlatform<?, ?, ?> getPlatform() {
+	static LoadContext getPlatform() {
+		int latestMajorVersion = 21; // Change this for Minecraft's major update
 		if (CommandAPI.getConfiguration().shouldUseLatestNMSVersion()) {
-			return new NMS_1_21_R1();
+			return new LoadContext(new NMS_1_21_R1(), () -> {
+				CommandAPI.logWarning("Loading the CommandAPI with a potentially incompatible NMS implementation.");
+				CommandAPI.logWarning("While you may find success with this, further updates might be necessary to fully support the version you are using.");
```
willkroboth, line 56: This warning and the one for loading a minor version could specify the situation. Something like `Forcefully loading the latest NMS version` and `Attempting to load NMS for the same minor version`. That could help identify why a certain platform was loaded when just looking at the log.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -68,8 +72,23 @@ public interface CommandAPIVersionHandler {
...
 				case "1.21", "1.21.1" -> new NMS_1_21_R1();
-				default -> throw new UnsupportedVersionException(version);
+				default -> null;
 			};
+			if (platform != null) {
+				return new LoadContext(platform, () -> {});
+			}
+			if (CommandAPI.getConfiguration().shouldBeLenientForMinorVersions()) {
+				int currentMajorVersion = Integer.parseInt(version.split("\\.")[1]);
+				if (latestMajorVersion == currentMajorVersion) {
```
willkroboth, line 82: Is it necessary to parse the version to an int? Would it be simpler to compare the version as a String?

```suggestion
				String currentMajorVersion = version.split("\\.")[1];
				if (currentMajorVersion.equals("21")) {
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-vh/src/main/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -68,8 +72,23 @@ public interface CommandAPIVersionHandler {
...
+					return new LoadContext(new NMS_1_21_R1(), () -> {
+						CommandAPI.logWarning("Loading the CommandAPI with a potentially incompatible NMS implementation.");
+						CommandAPI.logWarning("While you may find success with this, further updates might be necessary to fully support the version you are using.");
+					});
+				} else {
+					throw new UnsupportedVersionException(version);
+				}
```
willkroboth, line 89: Is this else block necessary? If it was removed, execution should continue to the `throw` statement currently on line 91, resulting in the same outcome.

```suggestion
				}
```

willkroboth (commented):
Just a few comments that don't affect functionality. Could be implemented or not. Otherwise looks good :+1:.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin-mojang-mapped/src/main/java/dev/jorel/commandapi/CommandAPIMain.java
```diff
@@ -51,7 +51,8 @@ public void onLoad() {
 			.missingExecutorImplementationMessage(fileConfig.getString("messages.missing-executor-implementation"))
 			.dispatcherFile(fileConfig.getBoolean("create-dispatcher-json") ? new File(getDataFolder(), "command_registration.json") : null)
 			.shouldHookPaperReload(fileConfig.getBoolean("hook-paper-reload"))
-			.skipReloadDatapacks(fileConfig.getBoolean("skip-initial-datapack-reload"));
+			.skipReloadDatapacks(fileConfig.getBoolean("skip-initial-datapack-reload"))
+			.beLenientForMinorVersions(fileConfig.getBoolean("lenient-for-minor-versions"));
```
willkroboth, line 55: The name of this option in the `config.yml` is now `be-lenient-for-minor-versions` (Same in `commandapi-bukkit-plugin`)
```suggestion
			.beLenientForMinorVersions(fileConfig.getBoolean("be-lenient-for-minor-versions"));
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/CommandAPIVersionHandler.java
```diff
@@ -43,12 +43,12 @@ private static boolean isMojangMapped() {
 		}
 	}
 	
-	static CommandAPIPlatform<?, ?, ?> getPlatform() {
+	static LoadContext getPlatform() {
 		if(profileId == null) {
 			System.out.println("Using default version 1.19.4");
-			return new MockNMS(new NMS_1_19_4_R3());
+			return new LoadContext(new MockNMS(new NMS_1_19_4_R3()), () -> {});
```
willkroboth, line 49: This could use the default `LoadContext` constructor
```suggestion
			return new LoadContext(new MockNMS(new NMS_1_19_4_R3()));
```

willkroboth (changes requested):

willkroboth (approved):

PR #596, opened 2024-08-20
Rework plugin config system

DerEchtePilz:
Just an idea I had, can be merged if good and wanted, but can also be closed.
The advantage of this system is that updating the config is possible. New values can be added and user configs that already exist get the new values. Furthermore, this system would also allow removing values if they are no longer needed, if that ever happens.


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/config/ConfigGenerator.java
```diff
@@ -0,0 +1,129 @@
...
+		for (String configOption : configOptions) {
+			ConfigurationSection section = config.getConfigurationSection(configOption);
+			if (section != null) {
+				sections.add(configOption);
+			}
+		}
+		for (String sectionName : sections) {
+			configOptions.remove(sectionName);
+		}
+		List<String> defaultConfigOptions = DefaultedBukkitConfig.ALL_OPTIONS.stream().map(option -> option.path()).toList();
```
JorelAli, line 119: - Don't use a stream here
- We're only ever using this to check if this collection contains a value, so we should be using a `Set` instead

DerEchtePilz: > - Don't use a stream here

Alright, rip the one-liner :smiling_face_with_tear:

DerEchtePilz:
Maybe Configurate could work, though I haven't used it before, only heard of it.
https://github.com/SpongePowered/Configurate

commandapi-core/src/main/java/dev/jorel/commandapi/config/DefaultedConfig.java
```diff
@@ -0,0 +1,64 @@
...
+		}, false
+	);
+
+	public static final CommentedConfigOption<String> MISSING_EXECUTOR_IMPLEMENTATION = new CommentedConfigOption<>(
+		new String[]{
+			"Missing executor implementation (default: \"This command has no implementations for %s\")",
+			"The message to display to senders when a command has no executor. Available",
+			"parameters are:",
+			"  %s - the executor class (lowercase)",
+			"  %S - the executor class (normal case)"
```
Strokkur424, line 36: `normal case`? I feel like this should be changed to `uppercase`

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/config/DefaultedBukkitConfig.java
```diff
@@ -0,0 +1,115 @@
...
+	public static final CommentedConfigOption<Boolean> SHOULD_HOOK_PAPER_RELOAD = new CommentedConfigOption<>(
+		new String[] {
+			"Hook into Paper's ServerResourcesReloadedEvent (default: true)",
+			"If \"true\", and the CommandAPI detects it is running on a Paper server, it will",
+			"hook into Paper's ServerResourcesReloadedEvent to detect when /minecraft:reload is run.",
+			"This allows the CommandAPI to automatically call its custom datapack-reloading",
+			"function which allows CommandAPI commands to be used in datapacks.",
+			"If you set this to false, CommandAPI commands may not work inside datapacks after",
+			"reloading datapacks."
+		}, false
```
Strokkur424, line 45: We are saying that this defaults to true. Then why is the value here set to false?

docssrc/src/config.md
```diff
@@ -10,7 +10,73 @@ The default `config.yml` is shown below:
...
+be-lenient-for-minor-versions: false
+
+# Hook into Paper's ServerResourcesReloadedEvent (default: true)
+# If "true", and the CommandAPI detects it is running on a Paper server, it will
+# hook into Paper's ServerResourcesReloadedEvent to detect when /minecraft:reload is run.
+# This allows the CommandAPI to automatically call its custom datapack-reloading
+# function which allows CommandAPI commands to be used in datapacks.
+# If you set this to false, CommandAPI commands may not work inside datapacks after
+# reloading datapacks.
+hook-paper-reload: false
```
Strokkur424, line 58: Same here. default: true, but the actual value is set to false.

Strokkur424 (changes requested):
I haven't really looked too deep into how the code works, but it looks good. I just highlighted some inconsistency

commandapi-core/src/main/java/dev/jorel/commandapi/config/DefaultedConfig.java
DerEchtePilz: No, it is like this in the config now. I mean, we can talk about changing the wording but this was copied over from the currently existing config.

Strokkur424: I mean, might as well change it whilst we are at it. That is up to you though

DerEchtePilz: I mean, the point of this system is that we could update stuff like this later too. For now, I think I'll leave it as is though.

Strokkur424: And the true/false inconsistencies?

DerEchtePilz: Those definitely need to be resolved before merging. While those won't be a problem for existing configs, new configs will get the wrong defaults which obviously shouldn't happen.

willkroboth: As a note, `%S` is not uppercase. It is the name of the class, so something like `CraftPlayer`: https://commandapi.jorel.dev/9.5.2/config.html#messages. Technically this is called PascalCase, though "normal case" sounds fine to me for this user-facing comment.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin-mojang-mapped/src/main/java/dev/jorel/commandapi/CommandAPIMain.java
```diff
@@ -133,4 +138,51 @@ private JavaPlugin getAndValidatePlugin(String pluginName) {
...
+
+	/**
+	 * In contrast to the superclasses method {@link org.bukkit.plugin.java.JavaPlugin#saveDefaultConfig()},
+	 * this doesn't fail silently if the config.yml already exists but instead will update the config with
+	 * new values if available.
+	 * <p>
+	 * This should fail silently if all values are set already.
+	 */
+	@Override
+	public void saveDefaultConfig() {
```
willkroboth, line 150: It looks like the logic in this method for checking if the config file exists and either creating or updating is shared with CommandAPIVelocity (and with the spigot-mapped plugin). I think this could be merged in `ConfigGenerator` using a method in `ConfigurationAdapter` that loads its platform-specific configuration object from a file.

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/config/VelocityConfigurationAdapter.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+@SuppressWarnings("ConfusingArgumentToVarargsMethod")
+public record VelocityConfigurationAdapter(YamlConfigurationLoader loader, CommentedConfigurationNode config) implements ConfigurationAdapter<ConfigurationNode, DefaultedVelocityConfig> {
+
+	@Override
+	public void setValue(String key, Object value) {
+		setValue0(key, value);
```
willkroboth, line 19: Why do `setValue` and `getValue` call `setValue0` and `getValue0`? Couldn't the code in `setValue0` just be moved to `setValue`?

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/config/VelocityConfigurationAdapter.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	@Override
+	public void setComment(String key, String[] comment) {
+		config.node(key.split("\\.")).comment(String.join("\n", comment));
```
willkroboth, line 24: I see the code `config.node(key.split("\\."))` repeated a lot. Should it be moved to its own method?

```java
@SuppressWarnings("ConfusingArgumentToVarargsMethod")
public CommentedConfigurationNode getNode(String path) {
	return config.node(path.split("\\."));
}
```

That would also mean that `ConfusingArgumentToVarargsMethod` would only need to be suppressed in one location (though I would like to figure out how to resolve that warning rather than suppressing it).

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/config/VelocityConfigurationAdapter.java
```diff
@@ -0,0 +1,105 @@
...
+	@Override
+	public Object getValue(String key) {
+		return getValue0(key);
+	}
+
+	@SuppressWarnings("DataFlowIssue")
+	@Override
+	public String[] getComment(String key) {
+		return config.node(key.split("\\.")).comment().split("\n");
+	}
```
willkroboth, line 36: It doesn't seem like this `DataFlowIssue` should be suppressed. If a node doesn't have a comment on it, then `.comment()` will return `null` and `.split("\n")` will throw an exception.

To reproduce, put the plugin on a Velocity server to get this config file (for some reason, the first generation doesn't have any comments, which seems like a separate issue).

image

If you then stop and restart the server, I got this error:

```
[17:23:29 ERROR]: Can't create plugin commandapi
com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) [Guice/ErrorInjectingConstructor]: NullPointerException: Cannot invoke "String.split(String)" because the return value of "org.spongepowered.configurate.CommentedConfigurationNode.comment()" is null
  at CommandAPIMain.<init>(CommandAPIMain.java:39)
  at VelocityPluginModule.configure(VelocityPluginModule.java:47)
  while locating CommandAPIMain

Learn more:
  https://github.com/google/guice/wiki/ERROR_INJECTING_CONSTRUCTOR

1 error

======================
Full classname legend:
======================
CommandAPIMain:       "dev.jorel.commandapi.CommandAPIMain"
VelocityPluginModule: "com.velocitypowered.proxy.plugin.loader.java.VelocityPluginModule"
========================
End of classname legend:
========================

        at com.google.inject.internal.InternalProvisionException.toProvisionException(InternalProvisionException.java:251) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.google.inject.internal.InjectorImpl$1.get(InjectorImpl.java:1151) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.google.inject.internal.InjectorImpl.getInstance(InjectorImpl.java:1186) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.velocitypowered.proxy.plugin.loader.java.JavaPluginLoader.createPlugin(JavaPluginLoader.java:131) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.velocitypowered.proxy.plugin.VelocityPluginManager.loadPlugins(VelocityPluginManager.java:168) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.velocitypowered.proxy.VelocityServer.loadPlugins(VelocityServer.java:362) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.velocitypowered.proxy.VelocityServer.start(VelocityServer.java:239) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.velocitypowered.proxy.Velocity.main(Velocity.java:71) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
Caused by: java.lang.NullPointerException: Cannot invoke "String.split(String)" because the return value of "org.spongepowered.configurate.CommentedConfigurationNode.comment()" is null
        at dev.jorel.commandapi.config.VelocityConfigurationAdapter.getComment(VelocityConfigurationAdapter.java:35) ~[?:?]
        at dev.jorel.commandapi.config.ConfigGenerator.generateWithNewValues(ConfigGenerator.java:52) ~[?:?]
        at dev.jorel.commandapi.CommandAPIMain.<init>(CommandAPIMain.java:67) ~[?:?]
        at dev.jorel.commandapi.CommandAPIMain$$FastClassByGuice$$a2b4c.GUICE$TRAMPOLINE(<generated>) ~[?:?]
        at dev.jorel.commandapi.CommandAPIMain$$FastClassByGuice$$a2b4c.apply(<generated>) ~[?:?]
        at com.google.inject.internal.DefaultConstructionProxyFactory$FastClassProxy.newInstance(DefaultConstructionProxyFactory.java:82) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.google.inject.internal.ConstructorInjector.provision(ConstructorInjector.java:114) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:91) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:300) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.google.inject.internal.ProviderToInternalFactoryAdapter.get(ProviderToInternalFactoryAdapter.java:40) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.google.inject.internal.SingletonScope$1.get(SingletonScope.java:169) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.google.inject.internal.InternalFactoryToProviderAdapter.get(InternalFactoryToProviderAdapter.java:45) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        at com.google.inject.internal.InjectorImpl$1.get(InjectorImpl.java:1148) ~[server.jar:3.3.0-SNAPSHOT (git-6100e675-b416)]
        ... 6 more
```

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/config/VelocityConfigurationAdapter.java
```diff
@@ -0,0 +1,105 @@
...
+
+	@Override
+	public Set<String> getKeys() {
+		Set<String> keys = new HashSet<>();
+		for (Object key : config.childrenMap().keySet()) {
+			keys.add((String) key);
+		}
+		return keys;
+	}
```
willkroboth, line 45: I believe we could directly return the `keySet` to achieve the same result.

```suggestion
	@Override
	public Set<String> getKeys() {
		return config.childrenMap().keySet();
	}
```

However, this current implementation seems flawed. On Bukkit, `ConfigurationAdapter#getKeys` returns all the options including those inside a configuration section. It seems to me that `childrenMap` returns the base level of the config node structure, which in this case would incorrectly include the key for `messages` and omit the key for `messages.missing-executor-implementation`.

willkroboth (changes requested):
Another miscellaneous thing I noticed while testing that didn't fit anywhere else. If I have a config that looks like this:

image

Once running the server the config gets updated to this:

image

This doesn't seem like a big deal since `- Essentials: null` seems to be interpreted the same, but I thought I'd bring it up.

Automated tests might've caught some of the things I brought up, and perhaps more that I missed.

commandapi-core/src/main/java/dev/jorel/commandapi/config/ConfigGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+	public static ConfigGenerator createNew(DefaultedConfig defaultedConfig) {
+		return new ConfigGenerator(defaultedConfig);
+	}
+
+	public <T, C extends DefaultedConfig> void populateDefaultConfig(ConfigurationAdapter<T, C> adapter) {
+		for (Map.Entry<String, CommentedConfigOption<?>> commentedConfigOption : defaultedConfig.getAllOptions().entrySet()) {
+			adapter.setValue(commentedConfigOption.getKey(), commentedConfigOption.getValue().option());
+			adapter.setComment(commentedConfigOption.getKey(), commentedConfigOption.getValue().comment());
+		}
+	}
```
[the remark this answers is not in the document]
DerEchtePilz: Yep, realized this already. Again, it isn't fully tested yet.

commandapi-core/src/main/java/dev/jorel/commandapi/config/ConfigGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+
+		boolean shouldRemoveValues = shouldRemoveOptions(existingConfig);
+
+		boolean wasConfigUpdated = false;
+		for (Map.Entry<String, CommentedConfigOption<?>> commentedConfigOption : defaultedConfig.getAllOptions().entrySet()) {
+			String path = commentedConfigOption.getKey();
+
+			// Update config option
+			if (existingConfig.contains(path)) {
+				updatedConfig.tryCreateSection(path, (C) defaultedConfig);
```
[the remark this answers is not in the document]
DerEchtePilz: Did you test that? Because there should be logic in place to handle comments updating.

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/config/VelocityConfigurationAdapter.java
DerEchtePilz: > for some reason, the first generation doesn't have any comments, which seems like a separate issue

The first generation? Does restarting the server (or rather the proxy) create the comments? Because I've brought this up on discord before, apparently Configurate's Yaml doesn't support comments and other formats might not be better for us.

commandapi-core/src/main/java/dev/jorel/commandapi/config/CommentedSection.java
```diff
@@ -0,0 +1,9 @@
+package dev.jorel.commandapi.config;
+
+import org.jetbrains.annotations.ApiStatus;
+
+import java.util.List;
+
+@ApiStatus.Internal
+record CommentedSection(List<String> comment) {
```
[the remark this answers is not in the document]
DerEchtePilz: Yeah, I am really not a fan of this to be honest. Maybe it would make more sense if we had more sections or even nested sections but I don't think we need this at this time.

commandapi-core/src/main/java/dev/jorel/commandapi/config/ConfigGenerator.java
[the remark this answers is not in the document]
willkroboth: Yea, I did run the plugin with the config above and it didn't update. Note though that if anything else is changed other than that comment, it will correctly detect that the config needs to be updated, in which case it will fix the comment.

There is code to detect if a `CommentedConfigOption` has the wrong comment:

https://github.com/JorelAli/CommandAPI/blob/bfea661c7c694ddcac088f47353efa31f57d3b83/commandapi-core/src/main/java/dev/jorel/commandapi/config/ConfigGenerator.java#L54-L56

However, the same does not happen for `CommentedSection` comments, so if those are the only thing that changes `wasConfigUpdated` will still be `false`.

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/config/VelocityConfigurationAdapter.java
willkroboth: Well, the first time there are no comments as pictured. The second time, I get that error and nothing updates because the code is interrupted. Sooo... I can't tell if restarting the server would add comments because the plugin crashes :P.

DerEchtePilz: Hmmm.. I wonder if we could just not use the CommentedConfigurationNode since comments are not really a thing here. Ideally we get support for that but for now I am thinking to leave out comments on Velocity.
So yeah, it obviously crashes when comments aren't present (which should be avoided).

willkroboth (changes requested):
Functionality seems good, so just some comments on code logic and organization.

I still think it would be good to get some unit tests to make sure all code paths and edge cases are verified.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-plugin-common/src/main/java/dev/jorel/commandapi/config/BukkitConfigurationAdapter.java
```diff
@@ -0,0 +1,158 @@
...
+		return this;
+	}
+
+	@Override
+	public ConfigurationAdapter<YamlConfiguration> createNew() {
+		return new BukkitConfigurationAdapter(new YamlConfiguration());
+	}
+
+	@Override
+	public void saveDefaultConfig(File directory, File configFile, Logger logger) {
```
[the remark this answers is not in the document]
DerEchtePilz: I know what you meant and I thought you might say this, however, the `ConfigGenerator` is only supposed to be used to generate/populate the config. Personally, I already think it's weird that the platform-specific adapters are responsible to save the config which is why I am definitely not willing to give that responsibililty to the `ConfigGenerator` class.

As for the exception handling, for Velocity I just copied what is there now and yeah, it probably should be changed to something similar to the way we handle exceptions on Bukkit.

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/config/VelocityConfigurationAdapter.java
```diff
@@ -0,0 +1,153 @@
...
+			return null;
+		}
+	}
+
+	@Override
+	public String[] getComment(String key) {
+		CommentedConfigOption<?> option = defaultVelocityConfig.getAllOptions().get(key);
+		CommentedSection section = defaultVelocityConfig.getAllSections().get(key);
+		return option != null ? option.comment() : section.comment();
+	}
```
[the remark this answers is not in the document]
DerEchtePilz: I mean, the `ConfigGenerator` is written with comments in mind. This right here is essentially telling it to ignore comments since the check within the generator that checks if the comments are different will always fail since the default comment is compared with itself.

commandapi-plugin/src/main/java/dev/jorel/commandapi/config/ConfigGenerator.java
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	public <T> void populateDefaultConfig(ConfigurationAdapter<T> adapter) {
+		for (Map.Entry<String, CommentedConfigOption<?>> commentedConfigOption : defaultConfig.getAllOptions().entrySet()) {
+			adapter.tryCreateSection(commentedConfigOption.getKey());
+			adapter.setValue(commentedConfigOption.getKey(), commentedConfigOption.getValue().option());
+			adapter.setComment(commentedConfigOption.getKey(), commentedConfigOption.getValue().comment());
+		}
+		adapter.complete();
+	}
```
willkroboth, line 26: Note that since `tryCreateSection` no longer sets the comments, when the CommandAPI directory is created for the first time and this code is run, the `messages` section does not have a comment:

image

So, basically this again https://github.com/JorelAli/CommandAPI/pull/596#discussion_r1770629791. I suppose you could add more code here to set the comments. Or, I think we could reduce the number of code paths that need to be tested and use `generateWithNewValues` with `existingConfig` being empty like I mentioned last time.

Automated unit tests would have caught this before merging :P

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/CommandAPIMain.java
```diff
@@ -35,24 +35,17 @@ public CommandAPIMain(ProxyServer server, Logger logger, @DataDirectory Path dat
...
+		VelocityConfigurationAdapter.createDummyInstance().saveDefaultConfig(configFile.getParent().toFile(), configFile.toFile(), logger);
 
 		// Load the file as a yaml node
 		ConfigurationNode configYAML;
 		try {
-			configYAML = YamlConfigurationLoader.builder().path(configFile).build().load();
+			YamlConfigurationLoader loader = YamlConfigurationLoader.builder()
+				.nodeStyle(NodeStyle.BLOCK)
+				.path(configFile)
+				.build();
```
willkroboth, line 47: Do we need to create the `YamlConfigurationLoader` again here? `VelocityConfigurationAdapter#saveDefaultConfig` creates an identical loader, which is accessible as a field of that record.

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/config/VelocityConfigurationAdapter.java
```diff
@@ -0,0 +1,156 @@
...
+	@Override
+	public void setComment(String key, String[] comment) {
+		StringBuilder commentBuilder = new StringBuilder();
+		for (int i = 0; i < comment.length; i++) {
+			if (i > 0) {
+				commentBuilder.append("\n");
+			}
+			commentBuilder.append(comment[i]);
+		}
+		node(key).comment(commentBuilder.toString());
```
willkroboth, line 42: Is this loop necessary? I believe we could use `String#join`:

```suggestion
		node(key).comment(String.join("\n", comment));
```

I think this is clearer for doing the same thing, and it cuts down the number of code paths that need to be tested by us.

willkroboth (commented):
Suprise merge lol, didn't realize we needed to merge this soon.

None of these are big issues, but I would have preferred a third opinion on a few conversations: https://github.com/JorelAli/CommandAPI/pull/596#discussion_r1770622729, https://github.com/JorelAli/CommandAPI/pull/596#discussion_r1789255235, https://github.com/JorelAli/CommandAPI/pull/596#discussion_r1789259068, and unit tests. When I have my own idea of how I would implement a system, its hard for me to tell if these are real concerns or just silly me problems :P.

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/config/VelocityConfigurationAdapter.java
DerEchtePilz: I think for this it really doesn't matter. Whenever I come across a "problem" like this my mind jumps straight to loops and I forget `String.join()` exists. I mean, `String.join()` would be a bit cleaner but I don't think it matters that much.

commandapi-platforms/commandapi-velocity/commandapi-velocity-plugin/src/main/java/dev/jorel/commandapi/CommandAPIMain.java
DerEchtePilz: > Do we need to create the `YamlConfigurationLoader` again here?

Yes. `saveDefaultConfig()` doesn't return anything so the loader used in that method is not accessible. Of course, we could change the return type to the `ConfigurationAdapter` itself, however I do not think it's worth it here because we would have to do the same on Bukkit (as the `saveDefaultConfig` method is inherited from `ConfigurationAdapter`) where it would be absolutely useless.

willkroboth: I suppose `CommandAPIMain` could create the `YamlConfigurationLoader` and pass it into `VelocityConfigurationAdapter#createDummyInstance`.

PR #615, opened 2024-11-25
fix documentation links

CJCrafter:
The current link, https://www.jorel.dev/CommandAPI/, gives a 404. Not sure if the link is outdated, server serving your website is down, etc...

Either way, I changed the 2 links to point towards the latest documentation.


JorelAli:
Thank you! We're going through a gradual domain transfer over the coming months, and recently moved the CommandAPI over to it's own organization.

PR #627, opened 2024-12-31
Fixes reloadDataPacks() on Paper 1.21.4 build 63 and above

DerEchtePilz:
Pointed out by @booky10 in Discord, due to the addition of the datapack lifecycle event in this commit (https://github.com/PaperMC/Paper/commit/feb8756567a3db1afcbbac465c8c6623ecfad1ea), `PackRepository#setSelected` now takes in an additional boolean. This results in a `NoSuchMethodException` when trying to run the current code.
This PR aims to fix that.
Also updates the current version to 9.7.1-SNAPSHOT in case we'd want to release this as a bug fix.

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
@@ -220,10 +220,9 @@ public class NMS_1_21_R3 extends NMS_Common {
 	private static final Field entitySelectorUsesSelector;
 	// private static final SafeVarHandle<ItemInput, CompoundTag> itemInput;
 	private static final Field serverFunctionLibraryDispatcher;
+	private static final MethodHandle minecraftServerSetSelected;
 	private static final boolean vanillaCommandDispatcherFieldExists;
 	private static final SafeVarHandle<MinecraftServer, FuelValues> minecraftServerFuelValues;
-
-	private static final MethodHandle minecraftServerSetSelected;
 
 	// Derived from net.minecraft.commands.Commands;
 	private static final CommandBuildContext COMMAND_BUILD_CONTEXT;
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
@@ -223,8 +223,7 @@ public class NMS_1_21_R3 extends NMS_Common {
 	private static final boolean vanillaCommandDispatcherFieldExists;
 	private static final SafeVarHandle<MinecraftServer, FuelValues> minecraftServerFuelValues;
 
-	private static MethodHandle minecraftServerSetSelected = null;
-	private static boolean setSelectedWithBoolean = false;
+	private static final MethodHandle minecraftServerSetSelected;
 
 	// Derived from net.minecraft.commands.Commands;
 	private static final CommandBuildContext COMMAND_BUILD_CONTEXT;
@@ -247,12 +246,14 @@ public class NMS_1_21_R3 extends NMS_Common {
 		serverFunctionLibraryDispatcher = CommandAPIHandler.getField(ServerFunctionLibrary.class, "h", "dispatcher");
 
 		MethodHandles.Lookup lookup = MethodHandles.lookup();
+		MethodHandle setSelected;
 		try {
-			minecraftServerSetSelected = lookup.findVirtual(PackRepository.class, "setSelected", MethodType.methodType(void.class, Collection.class, boolean.class));
-			setSelectedWithBoolean = true;
+			setSelected = lookup.findVirtual(PackRepository.class, "setSelected", MethodType.methodType(void.class, Collection.class, boolean.class));
 		} catch (NoSuchMethodException | IllegalAccessException e) {
-			// So, uhh... I guess we're on Paper 1.21.4 build 62 or earlier
+			// We're on Spigot or Paper 1.21.4 build 62 or earlier
+			setSelected = null;
 		}
+		minecraftServerSetSelected = setSelected;
 
 		boolean fieldExists;
 		try {
@@ -998,6 +999,10 @@ public class NMS_1_21_R3 extends NMS_Common {
 				}
 			}
 			return packResources;
+		}).exceptionally(exception -> {
+			CommandAPI.logException("Something went wrong while trying to collect resource packs!", exception);
+			// Return all currently selected packs
+			return this.<MinecraftServer>getMinecraftServer().getPackRepository().openAllSelected();
 		});
 
 		// Step 2: Convert all of the resource packs into ReloadableResources which
@@ -1018,6 +1023,10 @@ public class NMS_1_21_R3 extends NMS_Common {
 					LogUtils.getLogger().isDebugEnabled()).done();
 
 			return simpleReloadInstance.thenApply(x -> serverResources);
+		}).exceptionally(exception -> {
+			CommandAPI.logException("Something went wrong while trying to convert resource packs into ReloadableResources", exception);
+			// Return existing resources
+			return this.<MinecraftServer>getMinecraftServer().resources;
 		});
 
 		// Step 3: Actually load all of the resources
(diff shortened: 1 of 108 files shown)


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
```diff
@@ -1011,7 +1025,17 @@ public final void reloadDataPacks() {
 			this.<MinecraftServer>getMinecraftServer().resources.close();
 			this.<MinecraftServer>getMinecraftServer().resources = serverResources;
 			this.<MinecraftServer>getMinecraftServer().server.syncCommands();
-			this.<MinecraftServer>getMinecraftServer().getPackRepository().setSelected(collection);
+			if (!setSelectedWithBoolean) {
+				this.<MinecraftServer>getMinecraftServer().getPackRepository().setSelected(collection);
+			} else {
+				try {
+					minecraftServerSetSelected.invoke(this.<MinecraftServer>getMinecraftServer().getPackRepository(), collection, false);
```
booky10, line 1032:
```suggestion
					minecraftServerSetSelected.invoke(this.<MinecraftServer>getMinecraftServer().getPackRepository(), collection, true);
```

`true` would retain old behavior, but Idk wether you've checked that `false` also works

booky10 (approved):

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
DerEchtePilz: Ah, thanks! Didn't look into this further, but the old behaviour would probably be preferred.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
```diff
@@ -1011,7 +1025,17 @@ public final void reloadDataPacks() {
 			this.<MinecraftServer>getMinecraftServer().resources.close();
 			this.<MinecraftServer>getMinecraftServer().resources = serverResources;
 			this.<MinecraftServer>getMinecraftServer().server.syncCommands();
-			this.<MinecraftServer>getMinecraftServer().getPackRepository().setSelected(collection);
+			if (!setSelectedWithBoolean) {
```
willkroboth, line 1028: We could perhaps get rid of `setSelectedWithBoolean` and just do:

```suggestion
			if (minecraftServerSetSelected == null) {
```

If we did do that, it might be easier to read with the if statement flipped around. So rather than saying `if minecraftServerSetSelected does not exist (== null) then don't use it, otherwise alternative` we can have the logic `if minecraftServerSetSelected exists (!= null) then use it, otherwise alternative`.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
```diff
@@ -220,6 +223,9 @@ public class NMS_1_21_R3 extends NMS_Common {
 	private static final boolean vanillaCommandDispatcherFieldExists;
 	private static final SafeVarHandle<MinecraftServer, FuelValues> minecraftServerFuelValues;
 
+	private static MethodHandle minecraftServerSetSelected = null;
+	private static boolean setSelectedWithBoolean = false;
```
willkroboth, line 227: These fields could be final to match the others.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
```diff
@@ -240,6 +246,14 @@ public class NMS_1_21_R3 extends NMS_Common {
 		// For some reason, MethodHandles fails for this field, but Field works okay
 		serverFunctionLibraryDispatcher = CommandAPIHandler.getField(ServerFunctionLibrary.class, "h", "dispatcher");
 
+		MethodHandles.Lookup lookup = MethodHandles.lookup();
+		try {
+			minecraftServerSetSelected = lookup.findVirtual(PackRepository.class, "setSelected", MethodType.methodType(void.class, Collection.class, boolean.class));
+			setSelectedWithBoolean = true;
+		} catch (NoSuchMethodException | IllegalAccessException e) {
+			// So, uhh... I guess we're on Paper 1.21.4 build 62 or earlier
```
willkroboth, line 254: Adjusting this comment does not matter at all :P, but technically we do expect to sometimes be on Spigot

```suggestion
			// We're on Spigot or Paper 1.21.4 build 62 or earlier
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
```diff
@@ -1011,7 +1025,17 @@ public final void reloadDataPacks() {
...
-			this.<MinecraftServer>getMinecraftServer().getPackRepository().setSelected(collection);
+			if (!setSelectedWithBoolean) {
+				this.<MinecraftServer>getMinecraftServer().getPackRepository().setSelected(collection);
+			} else {
+				try {
+					minecraftServerSetSelected.invoke(this.<MinecraftServer>getMinecraftServer().getPackRepository(), collection, true);
+				} catch (Throwable e) {
+					for (StackTraceElement stackTraceElement : e.getStackTrace()) {
+						CommandAPI.logError(stackTraceElement.toString());
+					}
```
willkroboth, line 1036: The `CommandAPILogger` class does have the [`#severe(String message, Throwable exception)`](https://github.com/CommandAPI/CommandAPI/blob/37f96b59d000ce661629410af24624fc1d4f737f/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPILogger.java#L56) method which could be used here. That method is used by [`CommandAPIExecutor`](https://github.com/CommandAPI/CommandAPI/blob/37f96b59d000ce661629410af24624fc1d4f737f/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIExecutor.java#L81) twice, although it has to call `CommandAPI.getLogger()` since the `CommandAPI` class doesn't have a logging method that takes an exception. So maybe we'd also want to add a `CommandAPI#severe(String message, Throwable exception)` to use here and in `CommandAPIExecutor`?

willkroboth (commented):
Just static code analysis right now. None of these things are a big deal, functionality looks correct. Just "code style" stuff. 

I tried a bit to test this, but weirdly I couldn't reproduce datapacks not working with the old CommandAPI version? I was probably doing something wrong, so I'll try to figure that out again when I have the time.

willkroboth (approved):
Testing seems good - as far as I can tell, behavior seems the same as on old paper as well as consistent with Spigot. Although, I realized that I don't really know how the `reloadDataPacks` process is supposed to work :P. I was initially confused since the functions in my test datapack were reloading correctly. It seems that the functions get reloaded before `setSelected` is called, so even if that part errors, CommandAPI commands in datapacks still work. Fixing the exception is still definitely good since I'm sure the code after `setSelected` does important stuff. I just don't know what it does, so I don't know how to test it. But calling the method correctly does make sense to get us back to how it was working before, so I'll assume that's correct :+1:.

A few more comments, but nothing at all important. I'd be happy to merge this.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-nms/commandapi-bukkit-1.21.4/src/main/java/dev/jorel/commandapi/nms/NMS_1_21_R3.java
```diff
@@ -998,6 +999,10 @@ public final void reloadDataPacks() {
 				}
 			}
 			return packResources;
+		}).exceptionally(exception -> {
+			CommandAPI.logException("Something went wrong while trying to collect resource packs!", exception);
+			// Return all currently selected packs
+			return this.<MinecraftServer>getMinecraftServer().getPackRepository().openAllSelected();
```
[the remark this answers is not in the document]
DerEchtePilz: I was actually thinking the exact same thing.

PR #629, opened 2025-01-04
Register commands as plugin commands on Paper 1.20.6 and above

### LINKED ISSUE #578: Commands in the CommandMap don't have the correct permission on Paper
### CommandAPI version

9.5.1

### Minecraft version

1.20.5

### Are you shading the CommandAPI?

Yes

### What I did

I added this code to my `onEnable` and started the server:

```java
new CommandAPICommand("test")
        .withPermission("custom.permission")
        .executes(info -> {
            info.sender().sendMessage("Hello");
        })
        .register();

// Run later, since on Spigot commands are only moved once the server enables
//  On Paper you can access the command immediately
Bukkit.getScheduler().runTaskLater(
        this,
        // Spigot does not expose the CommandMap, though Paper does
        //  We can access it in both cases using the CommandAPI
        () -> getLogger().info(CommandAPIBukkit.get().getSimpleCommandMap().getCommand("test").getPermission()),
        0
);
```

### What actually happened

On a Paper 1.20.6 or 1.21 server, the log message says that the command's permission is `command.minecraft.test`.

### What should have happened

On a Paper 1.20.4 or Spigot server, the log message correctly reports that the permission is `custom.permission`.

### Server logs and CommandAPI config

_No response_

### Other

Note that the permission is correctly applied when running commands. This only affects the String returned when trying to see what the permission is via the CommandMap.

On Spigot, the CommandAPI has logic that makes sure the permissions of its commands in the CommandMap match up with the assigned Bukkit permissions, rather than keeping the default "vanilla-type" permission assigned when Bukkit moves commands from the Brigadier CommandDispatcher to its CommandMap:

https://github.com/JorelAli/CommandAPI/blob/d130bdee6647cb046fd326058b96cedfa56de59d/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/SpigotCommandRegistration.java#L154-L184

There is also [logic](https://github.com/JorelAli/CommandAPI/blob/d130bdee6647cb046fd326058b96cedfa56de59d/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/SpigotCommandRegistration.java#L237) when the CommandAPI registers commands after the server is enabled to correctly set the permission.

A similar thing does not immediately work on Paper. Paper's new Brigadier-centered command system allows accessing all registered CommandNodes as if they were Bukkit Commands through their BukkitBrigForwardingMap class:

```java
// Paper-Server, package io.papermc.paper.command.brigadier.bukkit
public class BukkitBrigForwardingMap extends HashMap<String, Command> {
    @Override
    public Command get(Object key) {
        CommandNode<?> node = this.getDispatcher().getRoot().getChild((String) key);
        if (node == null) {
            return null;
        }

        if (node instanceof BukkitCommandNode bukkitCommandNode) {
            return bukkitCommandNode.getBukkitCommand();
        }

        return PaperBrigadier.wrapNode(node);
    }
}
```

This special map looks up the requested node in the Brigadier dispatcher and usually constructs a new Bukkit Command instance. Since it creates a new Command object each time, setting the permission of the returned object doesn't keep that change for future calls.

### LINKED ISSUE #583: dispatchCommand not working for CommandAPI commands on 1.20.6+ Paper
### CommandAPI version

9.5.1

### Minecraft version

1.20

### Are you shading the CommandAPI?

No

### What I did

```
CommandAPICommand("shquests")
        .withSubcommand(
            CommandAPICommand("path")
                .withArguments(
                    EnumArgument("path", Path::class.java)
                )
                .executes(CommandExecutor { sender, args ->
                    val player = sender as Player

                    player.choosePath(theme, progressor, registry, args.getTyped("path"))
                }, ExecutorType.PLAYER)
        ).register();

// Sometime during runtime
Bukkit.dispatchCommand(me, 'shquests path TEST');
```

### What actually happened

image

### What should have happened

The command should have executed like it does when executed manually from a player (without `dispatchCommand`).

### Server logs and CommandAPI config

_No response_

### Other

_No response_

DerEchtePilz:
Fixes #583
Fixes #578 

The behaviour for #578 is unexpected, but corect. After this PR, instead of returning the wrong permission, `getCommandMap("commandname").getPermission()` will return `null`. This is the correct behaviour which can easily be verified if you follow the initialization chain starting here: https://github.com/PaperMC/Paper/blob/09f1f88f58a03f61092f8636ad780d42db87d8d9/paper-server/src/main/java/io/papermc/paper/command/brigadier/PaperBrigadier.java#L49

### DIFF
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
@@ -277,14 +277,10 @@ public abstract class CommandAPIBukkit<Source> implements CommandAPIPlatform<Arg
 
 	void updateHelpForCommands(List<RegisteredCommand> commands) {
 		Map<String, HelpTopic> helpTopicsToAdd = new HashMap<>();
-		Set<String> namespacedCommandNames = new HashSet<>();
 
 		for (RegisteredCommand command : commands) {
 			// Don't override the plugin help topic
 			String commandPrefix = generateCommandHelpPrefix(command.commandName());
-
-			// Namespaced commands shouldn't have a help topic, we should save the namespaced command name
-			namespacedCommandNames.add(generateCommandHelpPrefix(command.commandName(), command.namespace()));
 			
 			StringBuilder aliasSb = new StringBuilder();
 			final String shortDescription;
@@ -347,9 +343,6 @@ public abstract class CommandAPIBukkit<Source> implements CommandAPIPlatform<Arg
 					// Don't override the plugin help topic
 					commandPrefix = generateCommandHelpPrefix(alias);
 					helpTopic = generateHelpTopic(commandPrefix, shortDescription, currentAliasSb.toString().trim(), permission);
-
-					// Namespaced commands shouldn't have a help topic, we should save the namespaced alias name
-					namespacedCommandNames.add(generateCommandHelpPrefix(alias, command.namespace()));
 				}
 				helpTopicsToAdd.put(commandPrefix, helpTopic);
 			}
@@ -357,11 +350,6 @@ public abstract class CommandAPIBukkit<Source> implements CommandAPIPlatform<Arg
 
 		// We have to use helpTopics.put (instead of .addTopic) because we're overwriting an existing help topic, not adding a new help topic
 		getHelpMap().putAll(helpTopicsToAdd);
-
-		// We also have to remove help topics for namespaced command names
-		for (String namespacedCommandName : namespacedCommandNames) {
-			getHelpMap().remove(namespacedCommandName);
-		}
 	}
 
 	@Override
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
@@ -276,10 +276,11 @@ public abstract class CommandAPIBukkit<Source> implements CommandAPIPlatform<Arg
 	}
 
 	void updateHelpForCommands(List<RegisteredCommand> commands) {
-		if (getPaper().isPaperBrigAPI()) return;
 		Map<String, HelpTopic> helpTopicsToAdd = new HashMap<>();
 
 		for (RegisteredCommand command : commands) {
+			if (getPaper().isPaperBrigAPI() && !command.shouldGenerateHelpTopic()) continue;
+
 			// Don't override the plugin help topic
 			String commandPrefix = generateCommandHelpPrefix(command.commandName());
 			
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
@@ -276,6 +276,7 @@ public abstract class CommandAPIBukkit<Source> implements CommandAPIPlatform<Arg
 	}
 
 	void updateHelpForCommands(List<RegisteredCommand> commands) {
+		if (getPaper().isPaperBrigAPI()) return;
 		Map<String, HelpTopic> helpTopicsToAdd = new HashMap<>();
 
 		for (RegisteredCommand command : commands) {
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
@@ -11,6 +11,7 @@ import org.bukkit.help.HelpTopic;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Predicate;
@@ -35,21 +36,26 @@ public class PaperCommandRegistration<Source> extends CommandRegistrationStrateg
 
 	static {
 		CommandDispatcher<?> commandDispatcher;
-		// If we're using this class, we're operating on a Paper server with Paper's Brigadier API
-		// The class io.papermc.paper.command.brigadier.PluginCommandNode exists
 		Constructor<?> commandNode;
 		try {
-			commandNode = Class.forName("io.papermc.paper.command.brigadier.PluginCommandNode").getDeclaredConstructor(String.class, PluginMeta.class, LiteralCommandNode.class, String.class);
-
-			Class<?> paperCommands = Class.forName("io.papermc.paper.command.brigadier.PaperCommands");
-			Object paperCommandsInstance = paperCommands.getField("INSTANCE").get(null);
+			Object paperCommandsInstance = Class.forName("io.papermc.paper.command.brigadier.PaperCommands").getField("INSTANCE").get(null);
 			Field dispatcherField = Class.forName("io.papermc.paper.command.brigadier.PaperCommands").getDeclaredField("dispatcher");
 			dispatcherField.setAccessible(true);
 			commandDispatcher = (CommandDispatcher<?>) dispatcherField.get(paperCommandsInstance);
-		} catch (ClassNotFoundException | NoSuchMethodException | NoSuchFieldException | IllegalAccessException e) {
+		} catch (ClassNotFoundException | NoSuchFieldException | IllegalAccessException e) {
 			// This doesn't happen
-			commandNode = null;
 			commandDispatcher = null;
+		}
+
+		try {
+			commandNode = Class.forName("io.papermc.paper.command.brigadier.PluginCommandNode").getDeclaredConstructor(String.class, PluginMeta.class, LiteralCommandNode.class, String.class);
+		} catch (ClassNotFoundException | NoSuchMethodException e) {
+			try {
+				// If this happens, plugin commands on Paper are not identified with the PluginCommandNode anymore
+				commandNode = Class.forName("io.papermc.paper.command.brigadier.PluginCommandMeta").getDeclaredConstructor(PluginMeta.class, String.class, List.class);
+			} catch (ClassNotFoundException | NoSuchMethodException e1) {
+				commandNode = null;
+			}
 		}
 		staticDispatcher = commandDispatcher;
 		pluginCommandNodeConstructor = commandNode;
@@ -134,14 +140,33 @@ public class PaperCommandRegistration<Source> extends CommandRegistrationStrateg
 	@SuppressWarnings("unchecked")
 	private LiteralCommandNode<Source> asPluginCommand(LiteralCommandNode<Source> commandNode) {
 		try {
-			return (LiteralCommandNode<Source>) pluginCommandNodeConstructor.newInstance(
-				commandNode.getLiteral(),
-				CommandAPIBukkit.getConfiguration().getPlugin().getPluginMeta(),
-				commandNode,
-				getDescription(commandNode.getLiteral())
-			);
+			if (pluginCommandNodeConstructor.getDeclaringClass().getSimpleName().equals("PluginCommandNode")) {
+				return (LiteralCommandNode<Source>) pluginCommandNodeConstructor.newInstance(
+					commandNode.getLiteral(),
+					CommandAPIBukkit.getConfiguration().getPlugin().getPluginMeta(),
+					commandNode,
+					getDescription(commandNode.getLiteral())
+				);
+			} else {
+				setPluginCommandMeta(commandNode);
+				return commandNode;
+			}
 		} catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
 			throw new RuntimeException(e);
+		}
+	}
+
+	private void setPluginCommandMeta(LiteralCommandNode<Source> node) {
+		try {
+			Field metaField = node.getClass().getSuperclass().getDeclaredField("pluginCommandMeta");
+			metaField.setAccessible(true);
+			metaField.set(node, pluginCommandNodeConstructor.newInstance(
+				CommandAPIBukkit.getConfiguration().getPlugin().getPluginMeta(),
+				getDescription(node.getLiteral()),
+				Collections.emptyList()
+			));
+		} catch (NoSuchFieldException | InvocationTargetException | InstantiationException | IllegalAccessException e) {
+			// This doesn't happen
 		}
 	}
 
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
@@ -211,9 +211,7 @@ public class PaperCommandRegistration<Source> extends CommandRegistrationStrateg
 				if (helpTopic != null) {
 					description = ((HelpTopic) helpTopic).getShortText();
 				} else {
-					description = command.fullDescription()
-						.orElse(command.shortDescription()
-							.orElse("A command by the " + CommandAPIBukkit.getConfiguration().getPlugin().getName() + " plugin."));
+					description = command.shortDescription().orElse("A command by the " + CommandAPIBukkit.getConfiguration().getPlugin().getName() + " plugin.");
 				}
 				break;
 			}
--- a/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
+++ b/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
@@ -6,8 +6,6 @@ import com.mojang.brigadier.tree.CommandNode;
 import com.mojang.brigadier.tree.LiteralCommandNode;
 import com.mojang.brigadier.tree.RootCommandNode;
 import io.papermc.paper.plugin.configuration.PluginMeta;
-import net.kyori.adventure.text.Component;
-import org.bukkit.Bukkit;
 import org.bukkit.help.HelpTopic;
 
 import java.lang.reflect.Constructor;
@@ -15,10 +13,8 @@ import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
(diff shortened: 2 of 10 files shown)


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -89,9 +157,77 @@ public void unregister(String commandName, boolean unregisterNamespaces, boolean
...
+			metaField.setAccessible(true);
+			metaField.set(node, pluginCommandNodeConstructor.newInstance(
+				CommandAPIBukkit.getConfiguration().getPlugin().getPluginMeta(),
+				getDescription(node.getLiteral()),
+				getAliasesForCommand(node.getLiteral())
+			));
+		} catch (NoSuchFieldException | InvocationTargetException | InstantiationException | IllegalAccessException e) {
+			// This doesn't happen
+		}
+	}
```
willkroboth, line 199: We probably shouldn't be looking up the `pluginCommandMeta` field every time this method is called — we should cache this reflection as we do with other fields. Could perhaps use `SafeVarHandle` like so (although casting the generics is a little weird):

```java
private static final Constructor<Object> pluginCommandNodeConstructor;
private static final SafeVarHandle<CommandNode<?>, Object> metaField;

static {
	Constructor<?> commandNode;
	SafeVarHandle<CommandNode<?>, ?> metaFieldHandle = null;
	try {
		commandNode = Class.forName("io.papermc.paper.command.brigadier.PluginCommandNode").getDeclaredConstructor(String.class, PluginMeta.class, LiteralCommandNode.class, String.class);
	} catch (ClassNotFoundException | NoSuchMethodException e) {
		try {
			// If this happens, plugin commands on Paper are not identified with the PluginCommandNode anymore
			Class<?> pluginCommandMeta = Class.forName("io.papermc.paper.command.brigadier.PluginCommandMeta");
			commandNode = pluginCommandMeta.getDeclaredConstructor(PluginMeta.class, String.class, List.class);
			metaFieldHandle = SafeVarHandle.ofOrNull(CommandNode.class, "pluginCommandMeta", "pluginCommandMeta", pluginCommandMeta);
		} catch (ClassNotFoundException | NoSuchMethodException e1) {
			commandNode = null;
		}
	}
	pluginCommandNodeConstructor = (Constructor<Object>) commandNode;
	metaField = (SafeVarHandle<CommandNode<?>, Object>) metaFieldHandle;
}

private void setPluginCommandMeta(LiteralCommandNode<Source> node) {
	try {
		metaField.set(node, pluginCommandNodeConstructor.newInstance(
			CommandAPIBukkit.getConfiguration().getPlugin().getPluginMeta(),
			getDescription(node.getLiteral()),
			getAliasesForCommand(node.getLiteral())
		));
	} catch (InvocationTargetException | InstantiationException | IllegalAccessException e) {
		// This doesn't happen
	}
}
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -89,9 +157,77 @@ public void unregister(String commandName, boolean unregisterNamespaces, boolean
...
+			String namespaceStripped = "";
+			if (commandName.contains(":")) {
+				namespaceStripped = commandName.split(":")[1];
+			} else {
+				namespaceStripped = commandName;
+			}
+			if (command.commandName().equals(namespaceStripped) || Arrays.asList(command.aliases()).contains(namespaceStripped)) {
+				Object helpTopic = command.helpTopic().orElse(null);
+				if (helpTopic != null) {
+					description = ((HelpTopic) helpTopic).getShortText();
```
willkroboth, line 213: Not sure if this is the best way to use an `Optional`. Should probably use the fact that `command.helpTopic()` is an `Optional` rather than converting it to `null`.
```suggestion
				Optional<Object> helpTopic = command.helpTopic();
				if (helpTopic.isPresent()) {
					description = ((HelpTopic) helpTopic.get()).getShortText();
				} else {
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -89,9 +157,77 @@ public void unregister(String commandName, boolean unregisterNamespaces, boolean
...
+
+	private String getDescription(String commandName) {
+		String description = "";
+		for (RegisteredCommand command : CommandAPI.getRegisteredCommands()) {
+			String namespaceStripped = "";
+			if (commandName.contains(":")) {
+				namespaceStripped = commandName.split(":")[1];
+			} else {
+				namespaceStripped = commandName;
+			}
```
willkroboth, line 209: We don't need to compute `namespacedStripped` on every iteration of this loop. It only depends on `commandName`, which doesn't change.
```suggestion
		String namespaceStripped = "";
		if (commandName.contains(":")) {
			namespaceStripped = commandName.split(":")[1];
		} else {
			namespaceStripped = commandName;
		}
		for (RegisteredCommand command : CommandAPI.getRegisteredCommands()) {
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -89,9 +157,77 @@ public void unregister(String commandName, boolean unregisterNamespaces, boolean
...
+			if (command.commandName().equals(namespaceStripped) || Arrays.asList(command.aliases()).contains(namespaceStripped)) {
+				Object helpTopic = command.helpTopic().orElse(null);
+				if (helpTopic != null) {
+					description = ((HelpTopic) helpTopic).getShortText();
+				} else {
+					description = command.shortDescription().orElse("A command by the " + CommandAPIBukkit.getConfiguration().getPlugin().getName() + " plugin.");
+				}
+				break;
```
willkroboth, line 217: Is there a reason why `description` is its own variable? I believe we can simply return early here:
```suggestion
				if (helpTopic != null) {
					return ((HelpTopic) helpTopic).getShortText();
				} else {
					return command.shortDescription().orElse("A command by the " + CommandAPIBukkit.getConfiguration().getPlugin().getName() + " plugin.");
				}
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -89,9 +157,77 @@ public void unregister(String commandName, boolean unregisterNamespaces, boolean
...
+			}
+		}
+		return description;
+	}
+
+	private List<String> getAliasesForCommand(String commandName) {
```
willkroboth, line 223: `getDescription` strips the namespace from the `commandName`, but `getAliasesForCommand` doesn't do this pre-processing step. However, they both seem to do the same thing of looking for the `commandName` in the `RegisteredCommand` list. Is there a reason why these treat the `commandName` differently? Looking at when `ReigsteredCommand`s get created, it looks like they wouldn't ever have a namespace, so `getDescription`'s approach seems to be correct.

willkroboth (changes requested):
Just static code analysis right now, haven't re-tested the changes made since I last tried this.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -38,6 +100,11 @@ public boolean isBukkitCommand(CommandNode<Source> node) {
 		return isBukkitCommand.test(node);
 	}
 
+	@SuppressWarnings("unchecked")
+	public CommandDispatcher<Source> getPaperDispatcher() {
+		return (CommandDispatcher<Source>) getPaperDispatcher.get();
+	}
+
```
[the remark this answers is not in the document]
DerEchtePilz: I tried this and it doesn't seem to work. There is still a need for the normal, non-Paper dispatcher as the Paper one seemed to throw exceptions when trying to generate the dispatcher json I think. I'd definitely test this again, but there was a reason that there are both of them right now.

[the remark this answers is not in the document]
DerEchtePilz: Yep, tested it again, this is the error so both methods need to exist:
```
[11:59:23 ERROR]: Error occurred while enabling BugTestingCommandAPI v0.0.1 (Is it up to date?)
java.lang.UnsupportedOperationException: Cannot retrieve children from this node.
	at io.papermc.paper.command.brigadier.ShadowBrigNode.getChildren(ShadowBrigNode.java:19) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at net.minecraft.commands.synchronization.ArgumentUtils.serializeNodeToJson(ArgumentUtils.java:80) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at net.minecraft.commands.synchronization.ArgumentUtils.serializeNodeToJson(ArgumentUtils.java:81) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at bugtestingcommandapi-0.0.1-all.jar/io.github.derechtepilz.commandapi.nms.NMS_1_21_R3.createDispatcherFile(NMS_1_21_R3.java:432) ~[bugtestingcommandapi-0.0.1-all.jar:?]
	at bugtestingcommandapi-0.0.1-all.jar/io.github.derechtepilz.commandapi.CommandAPIHandler.writeDispatcherToFile(CommandAPIHandler.java:743) ~[bugtestingcommandapi-0.0.1-all.jar:?]
	at bugtestingcommandapi-0.0.1-all.jar/io.github.derechtepilz.commandapi.CommandAPIHandler.register(CommandAPIHandler.java:689) ~[bugtestingcommandapi-0.0.1-all.jar:?]
	at bugtestingcommandapi-0.0.1-all.jar/io.github.derechtepilz.commandapi.AbstractCommandAPICommand.register(AbstractCommandAPICommand.java:307) ~[bugtestingcommandapi-0.0.1-all.jar:?]
	at bugtestingcommandapi-0.0.1-all.jar/io.github.derechtepilz.commandapi.CommandAPICommand.register(CommandAPICommand.java:60) ~[bugtestingcommandapi-0.0.1-all.jar:?]
	at bugtestingcommandapi-0.0.1-all.jar/io.github.derechtepilz.commandapi.ExecutableCommand.register(ExecutableCommand.java:236) ~[bugtestingcommandapi-0.0.1-all.jar:?]
	at bugtestingcommandapi-0.0.1-all.jar/io.github.derechtepilz.bugtestingcommandapi.Main.onEnable(Main.java:41) ~[bugtestingcommandapi-0.0.1-all.jar:?]
	at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:280) ~[paper-api-1.21.4-R0.1-SNAPSHOT.jar:?]
	at io.papermc.paper.plugin.manager.PaperPluginInstanceManager.enablePlugin(PaperPluginInstanceManager.java:202) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at io.papermc.paper.plugin.manager.PaperPluginManagerImpl.enablePlugin(PaperPluginManagerImpl.java:109) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:520) ~[paper-api-1.21.4-R0.1-SNAPSHOT.jar:?]
	at org.bukkit.craftbukkit.CraftServer.enablePlugin(CraftServer.java:657) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at org.bukkit.craftbukkit.CraftServer.enablePlugins(CraftServer.java:606) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at net.minecraft.server.MinecraftServer.loadWorld0(MinecraftServer.java:743) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at net.minecraft.server.MinecraftServer.loadLevel(MinecraftServer.java:488) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at net.minecraft.server.dedicated.DedicatedServer.initServer(DedicatedServer.java:322) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1163) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
	at net.minecraft.server.MinecraftServer.lambda$spin$2(MinecraftServer.java:310) ~[paper-1.21.4.jar:1.21.4-183-f63dbea]
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
DerEchtePilz: Not sure if this is worth it though. I mean, we don't reduce code or make this more efficient so yeah, not so sure about this.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -17,17 +26,59 @@
...
+		try {
+			paperCommandsInstanceObject = Class.forName("io.papermc.paper.command.brigadier.PaperCommands").getField("INSTANCE").get(null);
+			dispatcherFieldObject = Class.forName("io.papermc.paper.command.brigadier.PaperCommands").getDeclaredField("dispatcher");
+		} catch (ReflectiveOperationException e) {
+			// Doesn't happen, or rather, shouldn't happen
+		}
+
+		paperCommandsInstance = paperCommandsInstanceObject;
+		dispatcherField = dispatcherFieldObject;
+		dispatcherField.setAccessible(true);
```
willkroboth, line 54: Could use `CommandAPIHandler#getField` here:
```suggestion
		try {
			Class<?> paperCommands = Class.forName("io.papermc.paper.command.brigadier.PaperCommands");
			paperCommandsInstanceObject = paperCommands.getField("INSTANCE").get(null);
			dispatcherFieldObject = CommandAPIHandler.getField(paperCommands, "dispatcher");
		} catch (ReflectiveOperationException e) {
			// Doesn't happen, or rather, shouldn't happen
		}

		paperCommandsInstance = paperCommandsInstanceObject;
		dispatcherField = dispatcherFieldObject;
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -38,6 +89,18 @@ public boolean isBukkitCommand(CommandNode<Source> node) {
...
+	public CommandDispatcher<Source> getPaperDispatcher() {
+		CommandDispatcher<?> commandDispatcher;
+		try {
+			commandDispatcher = (CommandDispatcher<?>) dispatcherField.get(paperCommandsInstance);
+		} catch (IllegalAccessException e) {
+			// This doesn't happen
+			commandDispatcher = null;
+		}
+		return (CommandDispatcher<Source>) commandDispatcher;
+	}
```
willkroboth, line 102: Could return a `CommandDispatcher<Source>` directly here:

```suggestion
	@SuppressWarnings("unchecked")
	public CommandDispatcher<Source> getPaperDispatcher() {
		try {
			return (CommandDispatcher<Source>) dispatcherField.get(paperCommandsInstance);
		} catch (IllegalAccessException e) {
			// This doesn't happen
			return null;
		}
	}
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -89,9 +154,79 @@ public void unregister(String commandName, boolean unregisterNamespaces, boolean
...
 			root.addChild(commandNode);
 		}
+		reloadHelpTopics.run();
+		CommandAPIBukkit.get().updateHelpForCommands(CommandAPI.getRegisteredCommands());
+	}
+
+	@SuppressWarnings("unchecked")
+	private LiteralCommandNode<Source> asPluginCommand(LiteralCommandNode<Source> commandNode) {
+		try {
+			if (pluginCommandNodeConstructor.getDeclaringClass().getSimpleName().equals("PluginCommandNode")) {
```
willkroboth, line 168: Could check `metaField` here:
```suggestion
			if (metaField == null) {
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/PaperCommandRegistration.java
```diff
@@ -89,9 +154,79 @@ public void unregister(String commandName, boolean unregisterNamespaces, boolean
...
+		try {
+			metaField.set(node, pluginCommandNodeConstructor.newInstance(
+				CommandAPIBukkit.getConfiguration().getPlugin().getPluginMeta(),
+				getDescription(node.getLiteral()),
+				getAliasesForCommand(node.getLiteral())
+			));
+		} catch (ReflectiveOperationException e) {
+			// This doesn't happen
+		}
+	}
```
willkroboth, line 194: Could only strip the namespace once:
```suggestion
	private void setPluginCommandMeta(LiteralCommandNode<Source> node) {
		String commandName = node.getName();
		if (commandName.contains(":")) {
			// Remove the namespace from the command name
			commandName = commandName.split(":")[1];
		}
		try {
			metaField.set(node, pluginCommandNodeConstructor.newInstance(
				CommandAPIBukkit.getConfiguration().getPlugin().getPluginMeta(),
				getDescription(commandName),
				getAliasesForCommand(commandName)
			));
		} catch (ReflectiveOperationException e) {
			// This doesn't happen
		}
	}
```

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandAPIBukkit.java
```diff
@@ -277,14 +277,12 @@ private String[] getUsageList(RegisteredCommand currentCommand) {
 
 	void updateHelpForCommands(List<RegisteredCommand> commands) {
 		Map<String, HelpTopic> helpTopicsToAdd = new HashMap<>();
-		Set<String> namespacedCommandNames = new HashSet<>();
 
 		for (RegisteredCommand command : commands) {
+			if (getPaper().isPaperBrigAPI() && !command.shouldGenerateHelpTopic()) continue;
```
willkroboth, line 282: Could call `getPaper().isPaperBrigAPI()` once, although that may be slightly less readable for not a big deal:
```suggestion
		boolean isPaperBrigAPI = getPaper().isPaperBrigAPI();
		for (RegisteredCommand command : commands) {
			if (isPaperBrigAPI && !command.shouldGenerateHelpTopic()) continue;
```

willkroboth (commented):
Redid the tests I mentioned [on Discord](https://discord.com/channels/745416925924032523/745419648970784821/1342127402524086303) and everything seems to be working well. Just a few optional comments, but functionality-wise seems good to merge.

PR #633, opened 2025-01-12
[Feat] AsyncOfflinePlayerArgument

Bram1903:
# Description
This pull request introduces the `AsyncOfflinePlayerArgument`, which extends the `SafeOverrideableArgument` class. The new class wraps the existing `OfflinePlayerArgument` within a `CompletableFuture<OfflinePlayer>`, enabling asynchronous handling of the offline player argument in commands. By using this approach, the logic involving API calls to fetch the offline player will be executed off the main thread, reducing the risk of blocking the server's main thread and improving performance.

## PR Checklist

- [x] Does the pull request have clear and descriptive commit messages that explain the changes made?
- [x] Does the CommandAPI compile successfully? (This will be verified automatically when the pull request is made.)
- [x] Has this been tested in a Minecraft server environment, including compatibility with the latest Minecraft version and Spigot or Paper?
- [ ] Are unit tests required for this feature, and if so, have they been added to the pull request?
- [x] Is documentation included in this pull request, and does it adequately explain the new functionality?

## Documentation
See https://github.com/CommandAPI/docs/pull/1

This PR is in continuation of #632, which I accidentally closed :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,10 @@ This is the current roadmap for the CommandAPI (as of 30th April 2024):
             <td valign="top"><b>9.7.1</b></td>
             <td valign="top">???</td>
             <td valign="top">
+				<b>CommandAPI Changes:</b>
+                <ul>
+                    <li>https://github.com/CommandAPI/CommandAPI/pull/633 Adds an <code>AsyncOfflinePlayerArgument</code> to allow asynchronous fetching of an offline player</li>
+                </ul>
                 <b>Bug Fixes:</b>
                 <ul>
                     <li>Fixes <code>PotionEffectArgument.NamespacedKey</code> not having suggestions in some versions</li>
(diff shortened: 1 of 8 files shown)


DerEchtePilz (approved):

willkroboth (approved):
Haven't personally tested functionality, but code looks good and makes sense to me.

I think it's fine to not include any automated tests in this PR given that `OfflinePlayerArgument` doesn't have any tests yet. I'm not sure how MockBukkit handles offline players, so that's something that could just be figured out later.

Bram1903:
The reason I didn't include a separate unit test for this is that the `OfflinePlayerArgument` is already being tested in the [ProfileArgumentTypeTests](https://github.com/CommandAPI/CommandAPI/blob/dev/dev/commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test-toolkit/src/test/java/dev/jorel/commandapi/arguments/ProfileArgumentTypeTests.java). This argument essentially wraps `OfflinePlayerArgument` into a `CompletableFuture<OfflinePlayerArgument>`, leveraging its existing functionality. Since the core behavior is already validated, additional tests here would be redundant.

DerEchtePilz:
Yeah, that doesn't test the actual OfflinePlayerArgument, that tests the custom implementation we have for the test toolkit to make it NMS independent.

Bram1903:
> Yeah, that doesn't test the actual OfflinePlayerArgument, that tests the custom implementation we have for the test toolkit to make it NMS independent.

Whoops, I see! Thanks for the heads up :-)

PR #636, opened 2025-03-12
Update the default namespace

DerEchtePilz:


commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/CommandNamespaceTests.java
```diff
@@ -51,7 +51,7 @@ Player enableWithNamespaces() {
 		SimpleCommandMap commandMap = mockPlatform.getSimpleCommandMap();
 		SpigotCommandRegistration<Object> spigotCommandRegistration = (SpigotCommandRegistration<Object>) mockPlatform.getCommandRegistrationStrategy();
 		for (CommandNode<Object> node : mockPlatform.getBrigadierDispatcher().getRoot().getChildren()) {
-			commandMap.register("minecraft", spigotCommandRegistration.wrapToVanillaCommandWrapper(node));
+			commandMap.register("commandapitest", spigotCommandRegistration.wrapToVanillaCommandWrapper(node));
```
willkroboth, line 54: I belive `minecraft` was used here to emulate what the Spigot server does when it transfers commands from the Brigadier dispatcher to the CommandMap. This part isn't related to the default namespace the CommandAPI uses, so it should stay as `minecraft`.

Some of the tests in this class were meant to check that the default namespace is used when a namespace isn't specified, so changing those instances to `commandapitest` makes sense. However, some tests (like `testSameCommandNameConflictWithDefaultNamespaceAndCustomNamespace` (oof that name is long)) were meant to cover the special logic that kicks in to handle edge cases with the `minecraft` namespace. Those tests should probably be updated to specify `minecraft` when previously they assumed it was the default, since we still want to cover that logic.

I think in general every `defaultNamespace` test is actually meant to test the special cases caused by the `minecraft` namespace, and logic for stuff like `commandapitest` are already covered by `customNamespace` things.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/CommandUnregisterTests.java
```diff
@@ -118,7 +172,7 @@ public void setUp() {
 		// Set up a Vanilla command
 		vanillaResults = Mut.of();
 		new CommandAPICommand("test")
-			.withAliases("minecraft:test")
+			.withAliases("commandapitest:test")
```
willkroboth, line 175: I think adding the namespaced version as an alias was a hack to get around the fact that the `minecraft` namespace isn't registered into the Brigadier dispatcher. Since other namespaces are put into the Brigadier dispatcher, we don't need to put it as an alias, and probably shouldn't since it causes the weird doubly-namespaced commands (`commandapitest:commandapitest:test`).

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/arguments/ArgumentMultiLiteralTests.java
```diff
@@ -86,6 +86,62 @@ void commandBuildingTestWithMultiLiteralArgument() {
...
+				              "executable": true
+				            },
+				            "option3": {
+				              "type": "literal",
+				              "executable": true
+				            }
+				          }
+				        }
+				      }
+				    },
```
willkroboth, line 144: We could perhaps specify the `minecraft` namespace when testing argument structure like here so that we don't get these duplicates listed in the json, which aren't relevant to what we're testing and just take up a lot of space. We could also just leave it; https://github.com/CommandAPI/CommandAPI/pull/602 modifies the json dump to indicate cases like this better, where an argument tree structure shows up in multiple places.

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-core/src/main/java/dev/jorel/commandapi/CommandRegistrationStrategy.java
```diff
@@ -7,6 +7,7 @@
 import com.mojang.brigadier.tree.RootCommandNode;
 import dev.jorel.commandapi.preprocessor.RequireField;
 
+import java.util.Arrays;
```
willkroboth, line 10: This import is unused

commandapi-platforms/commandapi-bukkit/commandapi-bukkit-test/commandapi-bukkit-test-tests/src/test/java/dev/jorel/commandapi/test/CommandNamespaceTests.java
```diff
@@ -535,7 +535,7 @@ public void testSameCommandNameConflictWithDefaultNamespaceAndCustomNamespace(bo
 				results.set("b");
 			});
 
-		defaultNamespace.register();
+		defaultNamespace.register("minecraft");
```
willkroboth, line 538: Maybe change this variable name to `minecraftNamespace` instead of `defaultNamespace`.

willkroboth (approved):

PR #642, opened 2025-04-21
Some updates to networking

willkroboth:
Some updates to the networking logic as suggested by @Timongcraft on Discord.

The `CommandAPI` Bukkit plugin now [provides](https://docs.papermc.io/paper/dev/plugin-yml/#provides) `CommandAPINetworking`. So, if a plugin depends on `CommandAPINetworking`, it will be able to load if `CommandAPINetworking` or `CommandAPI` is present. This makes sense, since `CommandAPI` provides the same networking support that `CommandAPINetworking` does. Provides also allows plugins to load classes from `CommandAPI` like they would from `CommandAPINetworking`, which *technically* doesn't apply here since `CommandAPINetworking` has some unique classes not present in `CommandAPI`. I don't currently expect anyone to load classes from `CommandAPINetworking` though, so this is not a problem right now.

`CommandAPI` Bukkit and `commandapi` Velocity both now have the `report-failed-packet-sends` config option. By default this is `true`, and the CommandAPI will throw a runtime exception if it cannot send a packet because the receiver doesn't have a new enough CommandAPI networking implementation. If set to `false`, these attempts will fail silently. Technically, Bukkit doesn't use this option since it doesn't send any packets right now, but it was easy enough to do this way with the config system, and it might be used eventually. Bukkit Networking doesn't have this option and will always throw since it also doesn't currently send any packets, and I didn't bother adding a config file to that.

Looking for feedback on whether `report-failed-packet-sends: false` should log a warning/error message instead of failing silently. It would be less intrusive than an unhandled exception, and it could still be turned off with `verbose-outputs` or `silent-logs`. (After typing that I kinda convinced myself a warning which you can turn off with `silent-logs` is the best option :P)

### DIFF
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/CommandAPIConfig.java
@@ -53,6 +53,8 @@ extends CommandAPIConfig<Impl>
 	boolean usePluginNamespace = false;
 	String namespace = null;
 
+	boolean reportFailedPacketSends;
+
 	/**
 	 * Sets verbose output logging for the CommandAPI if true.
 	 *
@@ -191,4 +193,14 @@ extends CommandAPIConfig<Impl>
 	 */
 	public abstract Impl usePluginNamespace();
 
+	/**
+	 * Sets whether the CommandAPI should throw exceptions if it cannot send a packet.
+	 *
+	 * @param value Whether an exception should be thrown when a packet cannot be sent
+	 * @return this CommandAPIConfig
+	 */
+	public Impl reportFailedPacketSends(boolean value) {
+		this.reportFailedPacketSends = value;
+		return instance();
+	}
 }
--- a/commandapi-core/src/main/java/dev/jorel/commandapi/InternalConfig.java
+++ b/commandapi-core/src/main/java/dev/jorel/commandapi/InternalConfig.java
@@ -58,6 +58,9 @@ public class InternalConfig {
 	// The default command namespace
 	private final String namespace;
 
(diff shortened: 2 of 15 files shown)


willkroboth:
I did convince myself and changed the config option to be `error-on-failed-packet-sends`. So when `true` it throws an error, and `false` logs a warning.

Timongcraft:
Why not have a option to disable the communication attempt directly instead of ignoring the error?

willkroboth:
`updateRequirements` can't do anything on Velocity if it can't communicate with the backend, so if it is expected to fail, I don't think there is any reason to try calling it. If you do assume it will always work, I think there should at least be a warning when a method does not do what it is supposed to. When `updateRequirements` fails, it can cause confusing desyncs where the client tells a player they can access a command when they really can't.

image

Timongcraft:
But I mean you need to disable it in the config so imo it's fine but if you think it should be liek this then...do that.

willkroboth:
No problem :) I wasn't exactly sure what you meant before

Timongcraft:
But why is it that the error is downgraded to a warning? I mean, if you set a config option to disable an exception I'd except it to be completely silenced. I understand that you still want a warning because something went wrong, but then why have a config option in the first place?

Timongcraft:
> The warning message can also be ignored by turning on the `silent-logs` option, so I thought this provided more possibilities.
I did not think of that. Well then I have nothing to say. Apologies

willkroboth:
:+1: No problem! Thanks for clarifying.

DerEchtePilz:
I just realized that this needs a docs PR to update the default config to include the new value.

PR #643, opened 2025-04-23
Integrate BytecodeAnalyser into the GitHub Actions build

DerEchtePilz:
Follow-up for #641 to now check compatibility for each commit so no accidental incompatible changes are made.


willkroboth (approved):
Makes sense and looks like it works. I wonder whether we could publish the BytecodeAnalyzer as a GitHub action, but that is definitely overkill for what we need.

I think it would be helpful if mappings issues caused the build to fail, so it would be more obvious when there is a problem. Right now you need to manually open the log and check if it found anything https://github.com/CommandAPI/CommandAPI/actions/runs/14626252087/job/41038318147#step:9:37. But that's a matter of how the BytecodeAnalyzer program works, so no reason to not merge this. I'm pretty sure if the Java program exits with a non-zero code this GitHub Action step will fail as well.

DerEchtePilz:
Yeah, if you look at the failing build right after the commit you linked to, you'll see that I added that there.
So yeah, that definitely exists since I really wanted it to fail the build if mapping issues occur.

PR #665, opened 2025-08-03
Support for individual method permission using annotations

wiktordudek:
This PR allows specifying command permissions individually for command methods, not only the entire command classes when using annotations. The `processCommand` method will skip `emitClassPermission` if method permission (`emitMethodPermission`) was found.

Until now, this code wouldn't work. Now it works.
```java
@Command("anvil")
@Permission("myplugin.anvil")
public class AnvilCommand {
    @Default
    public static void anvil(Player player) {
        player.openAnvil(null, true);
    }

    @Default
    @Permission("myplugin.anvil.other") // Important part - specify a different permission for this command only
    public static void anvil(Player player, @APlayerArgument Player target) {
        target.openAnvil(null, true);
    }
}
```


wiktordudek:
You're right, I didn't actually test the generated code and just relied on the "correctness" of the Google AutoService's output, hoping it was implemented correctly.

Thanks for pointing that out! 😅

Do you think replacing `null` with a negation to `CommandPermission.NONE` would solve the issue in this case instead of assigning `null` to `Argument.permission` by default?

So the "final" code would be:
```java
for (Argument argument : argumentsArray) {
 	if (argument.getArgumentPermission() != CommandPermission.NONE) {
 		argument.withPermission(meta.permission); 
 	}
}
```

DerEchtePilz (changes requested):
This PR needs to target the `dev/dev` branch.
Normally it would be enough to just change the base branch but that’s not possible here since `dev/dev` has a different module structure.
This change needs to be done in `commandapi-paper-annotations` and `commandapi-spigot-annotations` and also make sure that you are working on a branch that was created from the `dev/dev` branch since `master` and `dev/dev` have different commit histories.

PR #681, opened 2025-12-11
feat: add support for 1.21.11

Th0rgal:


Th0rgal:
@cursor review verbose=true

DerEchtePilz:
That's definitely not everything that would be required, but it's also not needed. We're done, the update is finished. If you want to test it, you can use our snapshot builds until we release.

PR #683, opened 2025-12-15
Add methods to register/unregister several commands at once

justlofe:
It'd be much better if user could register and unregister multiple commands at once due to long time execution of `Bukkit.reloadData()` (and so on, reloading after each command registered is very expensive). I've seen a code attempting to do so by delaying call to `reloadData()`, but this helps only for unregistering.

So, what i've done:
- moved main register logic from `AbstractCommandAPICommand` to `CommandAPI` (because why the `unregister` method located in `CommandAPI` class, and `register` is not?)
- added methods to register (and unregister) multiple commands at once
- made unregistered commands can be registered again

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,9 @@
 		<!-- Outputs -->
 		<module>commandapi-annotations</module>
 		<module>commandapi-testing</module>
+
+		<!-- Code coverage reports -->
+		<module>commandapi-codecov</module>
 	</modules>
 
 	<dependencyManagement>


DerEchtePilz:
In case you don't shade, make sure you downloaded the plugin from the latest `dev/dev` GitHub Actions build.

DerEchtePilz:
Yes, not running asynchronous code in a space that for the most part isn't thread-safe is probably a good idea 👍 
Generally though, in this space it is usually expected that things are not thread safe, and given how there is definitely lots of not thread-safe code in the CommandAPI this also applies here.
Happy to hear you found a working solution though.

On a different note, if you want to open another PR that addresses your initial goal of adding a method that can unregister several commands at once you are welcome to do that. Could probably even expose the `UnregisterInformation` for that.
However, and you can correct me on that if you see it differently, I don't really see a use in a method that can register several commands at once. Given how you have to construct the `CommandAPICommand` or `CommandTree` object anyway, I don't really see how useful such a method really is.

PR #691, opened 2026-04-07
Switch build system to Gradle

DerEchtePilz:
Haven't figured out how publishing works, but this should create identical jars.

The most noticable difference is probably the build process: instead of always specifiying a profile, that is now only required when not compiling something.
To build everything:
```
./gradlew build
```

To build the equivalent of the `Platform.Paper` profile:
```
./gradlew build -Pprofile=paper
```

Publishing to maven local works by supplying the `publishToMavenLocal` task alongside the `build` task.


willkroboth:
I got the SpecialSource remapping working, so the Spigot plugin jar is correct now. It looks like all our plugin jars are identical (except that they are no longer filled with Maven metadata), and the example projects work, so I think this is good to merge!

PR #694, opened 2026-06-20
Dev/26.2

willkroboth:
Support for Minecraft 26.2.  Changes that affected us were `ColorArgument -> TeamColorArgument` and `net.minecraft.advancements.criterion.MinMaxBounds -> net.minecraft.advancements.predicates.MinMaxBounds` (ChatColorArgument and Integer/DoubleRangeArgument).

Bukkit and Spigot NMS got a `26-common` version for shared logic between 26.1 and 26.2. For future versions, this will probably expand to be like `NMS_Common`. For reference, we aren't building off `NMS_Common` anymore since it ends up being Spigot-mapped for compatibility with older versions, while `NMS_26_Common` ends up Mojang-mapped in the Spigot CommandAPI jar.

Note that Paper is currently set to be the version `26.2.build.25-alpha`. Even though it is an alpha version, I believe this will continue to be available, so we don't *have* to update it to a "stable" release version. I believe you can set it up as a version range to always get the latest version. Locally that might be a little annoying given that we're using Maven, not Gradle, so I think you'd have to manually re-run `./setupPaperNMS` if the resolved version changes. But, something to think about for the Gradle build workflow.


DerEchtePilz:
Yeah, talking about the Gradle stuff, I believe it would be good to merge that first and then update this PR to contain new Gradle modules as well since I most definitely don’t have the time to update that branch any time soon.

willkroboth:
Set up Gradle and made the Paper version a `+` wildcard thing so we get the latest automatically.

PR #698, opened 2026-08-09
Fix registerCommand(Class<?>) with isolated class loaders

TheKinrar:
The new Paper variant of CommandAPI is a Paper plugin (with the paper-plugin.yml definition) so it gets an isolated class loader that cannot load classes from other plugins. Therefore it will fail to load a class generated with the annotation processor (unless CommandAPI is shaded).

This simply uses the already existing class loader to get the class locally, which is easy as we're already getting a Class, not only a name.


DerEchtePilz:
Uhh, I guess there's nothing wrong with this, however, if you depend on the CommandAPI plugin, shouldn't Paper put your plugin and the CommandAPI plugin on the same classpath and should therefore be able to load the classes just fine?

TheKinrar:
It's one-way only. Your plugin will be able to load CommandAPI classes, not the other way around

